PR #29, opened 2022-10-04
Implement subset of API call recording for scene reconstruction [v11]

prybicki:


prybicki:
Thank you @Jakub-Krakowiak for this work! From now, I am taking over the responsibility of finishing the implementation to register other (non-scene) calls. I moved the code to `feature/tape` and will make a new PR once the feature is completed.

PR #40, opened 2022-11-02
Remove managed memory to support v11 on Windows

prybicki:
- This is meant only as a quick surgery to enable blocked tasks
- Some optimizations and improvements will be done in the future

### DIFF
--- a/include/VArrayProxy.hpp
+++ b/include/VArrayProxy.hpp
@@ -21,7 +21,7 @@ struct VArrayProxy
 	VArray::Ptr untyped() { return src; };
 	VArray::ConstPtr untyped() const { return src; };
 
-	void copyFrom(const T* srcRaw, std::size_t count) { src->setData(srcRaw, count); }
+	void setData(const T* srcRaw, std::size_t count) { src->setData(srcRaw, count); }
 
 	std::size_t getCount() const { return src->getElemCount(); }
 	std::size_t getCapacity() const { return src->getElemCapacity(); }
--- a/include/VArrayProxy.hpp
+++ b/include/VArrayProxy.hpp
@@ -31,11 +31,11 @@ struct VArrayProxy
 	std::size_t getBytesInUse() const { return src->elemCount * sizeof(T); }
 	void resize(std::size_t newCount, bool zeroInit=true, bool preserveData=true) { src->resize(newCount, zeroInit, preserveData); }
 
-	T*          getHostPtr()           { return reinterpret_cast<T*>(src->getWritePtr(MemLoc::host())); }
-	const T*    getHostPtr()     const { return reinterpret_cast<const T*>(src->getReadPtr(MemLoc::device())); }
-	T*          getDevicePtr()         { return reinterpret_cast<T*>(src->getWritePtr(MemLoc::device())); }
-	const T*    getDevicePtr()   const { return reinterpret_cast<const T*>(src->getReadPtr(MemLoc::device())); }
-	CUdeviceptr getCUdeviceptr() const { return reinterpret_cast<CUdeviceptr>(src->getWritePtr(MemLoc::device())); }
+	T*          getHostPtr()           { return reinterpret_cast<T*>(src->getWritePtr(MemLoc::Host)); }
+	const T*    getHostPtr()     const { return reinterpret_cast<const T*>(src->getReadPtr(MemLoc::Device)); }
+	T*          getDevicePtr()         { return reinterpret_cast<T*>(src->getWritePtr(MemLoc::Device)); }
+	const T*    getDevicePtr()   const { return reinterpret_cast<const T*>(src->getReadPtr(MemLoc::Device)); }
+	CUdeviceptr getCUdeviceptr() const { return reinterpret_cast<CUdeviceptr>(src->getWritePtr(MemLoc::Device)); }
 
 	T&       operator[](int idx)       { return (reinterpret_cast<T*>(src->managedData))[idx]; }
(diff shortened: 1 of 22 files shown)


src/api/api.cpp
```diff
@@ -349,7 +349,7 @@ rgl_graph_get_result_data(rgl_node_t node, rgl_field_t field, void* data)
 		}
```
msz-rai, line 349: Since FormatPointsNode is `IPointsNode` if statement above is not needed. Could be simplified to:
```suggestion
auto pointCloudNode = Node::validatePtr<IPointsNode>(node);
VArray::ConstPtr output = pointCloudNode->getFieldData(field, nullptr);
		}
```

src/VArray.cpp
```diff
@@ -2,91 +2,115 @@
...
+	}
 }
 
-VArray::VArray(VArray &&src)
-: typeInfo(src.typeInfo)
-, sizeOfType(src.sizeOfType)
-, managedData(src.managedData)
-, elemCapacity(src.elemCapacity)
-, elemCount(src.elemCount)
+void VArray::migrateToLocation(MemLoc newLoc)
```
msz-rai, line 110: Return if `newLoc == currentLocation`

src/VArray.cpp
```diff
@@ -2,91 +2,115 @@
...
 {
-	CHECK_CUDA(cudaMallocManaged(&managedData, elemCapacity * sizeOfType));
-	CHECK_CUDA(cudaMemcpy(managedData, other.managedData, elemCapacity * sizeOfType, cudaMemcpyDefault));
+	MemLoc location = locationHint.has_value() ? locationHint.value() : currentLocation;
+	if (location == MemLoc::Host) {
+		CHECK_CUDA(cudaFreeHost(ptr));
+	}
+	if (location == MemLoc::Device) {
+		CHECK_CUDA(cudaFree(ptr));
+	}
```
msz-rai, line 107: Set `ptr` to `nullptr` just in case :)

include/graph/Nodes.hpp
```diff
@@ -42,6 +42,7 @@ struct FormatPointsNode : Node, IPointCloudDescription
 
 	static void formatAsync(const VArray::Ptr& output, const IPointsNode::Ptr& input,
 	                        const std::vector<rgl_field_t>& fields, cudaStream_t stream);
+	static VArrayProxy<GPUFieldDesc>::Ptr getGPUFields(IPointsNode::Ptr input, const std::vector<rgl_field_t>& fields, cudaStream_t stream);
 
 	inline VArray::ConstPtr getData() const { return output; }
```
msz-rai, line 47: Could be removed since we have `getFieldData()`?

include/VArrayProxy.hpp
```diff
@@ -18,29 +18,28 @@ struct VArrayProxy
...
 	{ return VArrayProxy<T>::ConstPtr(new VArrayProxy<T>(std::const_pointer_cast<VArray>(src))); }
 
-	VArrayProxy<T>::Ptr clone() const
-	{ return VArrayProxy<T>::create(src->clone()); }
-
 	VArray::Ptr untyped() { return src; };
 	VArray::ConstPtr untyped() const { return src; };
 
-	void copyFrom(const T* srcRaw, std::size_t count) { src->copyFrom(srcRaw, count); }
+	void copyFrom(const T* srcRaw, std::size_t count) { src->setData(srcRaw, count); }
```
msz-rai, line 24: I'd suggest having the same names of functions for both - VArray and VArrayProxy.

msz-rai (changes requested):
Solid re-implementation! Please take look at the comments.

test/src/VArrayTest.cpp
```diff
@@ -0,0 +1,47 @@
...
+	VArrayProxy<int>::Ptr array = VArrayProxy<int>::create(1);
+	int value = 4;
+
+	{
+		// Goal:
+		// CPU: ||
+		// GPU: |4|
+		int *devicePtr = array->getWritePtr(MemLoc::Device);
+		EXPECT_TRUE(devicePtr != nullptr);
+		CHECK_CUDA(cudaMemcpy(devicePtr, &value, sizeof(int), cudaMemcpyDefault));
```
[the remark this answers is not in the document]
prybicki: Did you mean "let's test it additionally" or "you could have used it here"? My point here was to test `getWritePtr`.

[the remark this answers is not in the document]
msz-rai: "you could have used it here" because VArrayProxy has the same function as written, and it is worth testing it.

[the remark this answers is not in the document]
prybicki: This code sequence (with `getWritePtr`) is a bit different than using `setData`. The latter would set data on the CPU if it were the current location. I meant to set precisely GPU data here. I will happily test the full API of VArray once we decide on the design. For now, I'm leaving as-is, pardon my laziness ;)

src/api/api.cpp
prybicki: Good point. The suggested change seems wrong but I applied it manually.

include/graph/Nodes.hpp
prybicki: No, these two have different purposes. `getGPUFields()` is a helper function preparing instructions on how to format on the GPU. `getFieldData()` just returns the output.

msz-rai: I was referring to the function `getData()` :smiley:

prybicki: Ah right. Removed ;)

msz-rai (approved):
Thank you for applying the changes. After tests, I can confirm that problem on Windows has been fixed.

PR #42, opened 2022-11-04
Fix crash on Windows in GAS update [To Be Verified]

prybicki:


include/scene/Mesh.hpp
```diff
@@ -38,6 +38,8 @@ struct Mesh : APIObject<Mesh>
 
 	// Shared between buildGAS() and updateGAS()
 	OptixBuildInput buildInput;
+		CUdeviceptr vertexBuffers[1];
```
msz-rai, line 41: Invalid indent

msz-rai (approved):
Crash has been fixed! Thanks

PR #45, opened 2022-11-07
Move private headers to src directory

prybicki:


Jakub-Krakowiak:
in core.h we need RGL_API macro, that was in include/macros/visibility - now in src/..., I propose moving this macro inside core.h

alek-kam-robotec-ai (changes requested):
When compiling using a .so file alongside headers under the include directory, build exits with the following error message:
/home/alekkam/Documents/code/repos/dev/RobotecGPULidar/include/rgl/api/core.h:3:10: fatal error: 'macros/visibility.h' file not found
#include <macros/visibility.h>
         ^~~~~~~~~~~~~~~~~~~~~
1 error generated.

prybicki:
Thanks for your feedback. I will prepare a fix in a moment.

prybicki:
@Jakub-Krakowiak @alek-kam-robotec-ai I've moved the macro into core.h. It compiles on my machine. Could you please check again if it works for you as well?

alek-kam-robotec-ai (approved):

PR #46, opened 2022-11-07
Log build info (branch and commit SHA1)

prybicki:


msz-rai (approved):
Good helper for debugging! I have tested on Windows as I was asked:
image

PR #50, opened 2022-11-17
Simplify nodes interface

prybicki:
@msz-rai `Nodes.hpp` changes are best to read fully instead of diff.


msz-rai (approved):
Very clean and simple! Well done :+1:

PR #55, opened 2022-11-22
Fix error when raytracing on empty scene

prybicki:


src/gpu/optixPrograms.cu
```diff
@@ -53,17 +52,11 @@ Vec3f decodePayloadVec3f(const Vec3fPayload& src)
...
 __forceinline__ __device__
-void saveRayResult(Vec3f* xyz=nullptr)
+void saveRayResult(const Vec3f* xyz=nullptr, const Vec3f* origin=nullptr)
 {
 	const int rayIdx = optixGetLaunchIndex().x;
-	Vec3f origin = decodePayloadVec3f({
-		optixGetPayload_0(),
-		optixGetPayload_1(),
-		optixGetPayload_2()
-	});
```
prybicki, line 66: `optixGetPayload_0` is not available when called from `__raygen__` function, hence it was moved to the callee side.

src/gpu/optixPrograms.cu
```diff
@@ -53,21 +52,14 @@ Vec3f decodePayloadVec3f(const Vec3fPayload& src)
...
-
 	if (ctx.xyz != nullptr) {
 		// Return actual XYZ of the hit point or infinity vector with signs of the ray.
```
msz-rai, line 61:
```suggestion
		// Return actual XYZ of the hit point or infinity vector.
```

msz-rai (approved):
LGTM!

PR #66, opened 2022-11-30
Improve lazy initialization control flow, fix AutoTape and Log CMake-provided path

prybicki:


src/api/api.cpp
```diff
@@ -135,6 +123,26 @@ static rgl_status_t rglSafeCall(Fn fn)
...
+	static bool initCalled = false;
+	if (initCalled) {
+		return;
+	}
+	initCalled = true;
+	rgl_status_t initStatus = rglSafeCall([&]() {
+		Logger::getOrCreate();
+		Optix::getOrCreate();
+		if (isCompiledWithAutoTape()) {
+			tapeRecord.emplace(std::filesystem::path(RGL_AUTO_TAPE_PATH));
```
msz-rai, line 137: I'd leave log info that auto tape has started on {path}

msz-rai (approved):
LGTM

src/api/api.cpp
prybicki: Good idea!

PR #84, opened 2023-01-16
Introduce graph execution structure

prybicki:


src/graph/Graph.cpp
```diff
@@ -106,16 +95,73 @@ void runGraph(Node::Ptr userNode)
...
-		Node::Ptr node = *graph.begin();
-		RGL_DEBUG("Destroying node {}", (void*) node.get());
-		graph.erase(node);
-		node->inputs.clear();
-		node->outputs.clear();
-		Node::release(node.get());
+	if (!preserveNodes) {
+		std::set<std::shared_ptr<Node>> graphNodes = anyNode->hasGraph()
+		                                           ? anyNode->getGraph()->getNodes()
+												   : Graph::findConnectedNodes(anyNode);
```
msz-rai, line 103: wrong indent

src/graph/Graph.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+	void run();
+	const std::set<std::shared_ptr<Node>>& getNodes() const { return nodes; }
+	const std::vector<std::shared_ptr<Node>>& getExecutionOrder();
+
+	virtual ~Graph();
+private:
+	Graph() { stream = std::make_shared<CudaStream>(); }
+
+	static std::vector<std::shared_ptr<Node>> findExecutionOrder(std::set<std::shared_ptr<Node>> nodes);
```
msz-rai, line 40:
```suggestion
	std::vector<std::shared_ptr<Node>> findExecutionOrder();
```
No need to be static

msz-rai (approved):
LGTM. Pair programming did the job. I fully understand the conceptual decisions taken in this PR.
I have left two comments to consider

src/graph/Graph.cpp
prybicki: Fixed

src/graph/Graph.hpp
prybicki: No need to be a member too. IMHO static underscores the fact that it will not modify cached execution order. Leaving as-is.

msz-rai: Ok, it makes sense :+1:

PR #85, opened 2023-01-17
[v14] Add GraphStress.Async test

prybicki:


test/src/graphStressTest.cpp
```diff
@@ -0,0 +1,243 @@
...
+static std::mt19937 randomGenerator {randomSeed};
+
+// randomizeIndices(10,3): 9, 0, 5
+std::vector<int> randomizeIndices(int containerSize, int firstN=0) {
+	std::vector<int> indices;
+	indices.resize(containerSize);
+	std::iota(indices.begin(), indices.end(), 0);
+	std::shuffle(indices.begin(), indices.end(), randomGenerator);
+	if (firstN > 0) {
+		indices.resize(containerSize);
```
msz-rai, line 30:
```suggestion
		indices.resize(firstN);
```

msz-rai (dismissed):
Well-designed async test. It compiles and passes on my machine. However, I have found a mistake in `randomizeIndices` function and addressed some comments.

msz-rai:
When you resolve conversations, please rebase to `v13` branch.

src/graph/Node.hpp
```diff
@@ -54,6 +54,7 @@ struct Node : APIObject<Node>, std::enable_shared_from_this<Node>
 	bool hasGraph() { return graph.lock() != nullptr; }
 
 	std::string getName() const { return name(typeid(*this)); }
+	virtual std::string getArgsString() const { return {}; }
```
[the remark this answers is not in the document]
prybicki: [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it). Added a comment saying it's optional. Implementing this completely is substantial work, so I decided to refrain from smuggling it into this PR.

test/src/graphStressTest.cpp
prybicki: Great catch, applied manually.

prybicki:
Rebased on v13. Adding @PiotrMrozikRobotec to review since he's working on the graph tests.

test/src/asyncStressTest.cpp
```diff
@@ -0,0 +1,242 @@
...
+	// Config
+	int GRAPH_COUNT = 16;  // Number of graphs executed in parallel
+	const int GRAPH_SIZE = 1024;  // Number of TransformPointsNodes in each graph
+	const int GRAPH_QUERIES = 32;  // How many nodes are checked for results (should be a subset)
+	const int GRAPH_EPOCHS = 32;  // How many times all graphs are executed, changed and queried
+	const float EPSILON = 2 * 1E-6;  // Tolerance for GPU-CPU matrix multiplication differences
+	ASSERT_THAT(GRAPH_QUERIES, Lt(GRAPH_SIZE));
+
+	// Optional logging
+	// EXPECT_RGL_SUCCESS(rgl_configure_logging(RGL_LOG_LEVEL_ALL, nullptr, true));
```
PiotrMrozikRobotec, line 152: I would prefer some local DEBUG_WITH_LOGS macro in the file, instead of commented code.

test/src/asyncStressTest.cpp
```diff
@@ -0,0 +1,242 @@
...
+{
+	using namespace testing;
+
+	// Config
+	int GRAPH_COUNT = 16;  // Number of graphs executed in parallel
+	const int GRAPH_SIZE = 1024;  // Number of TransformPointsNodes in each graph
+	const int GRAPH_QUERIES = 32;  // How many nodes are checked for results (should be a subset)
+	const int GRAPH_EPOCHS = 32;  // How many times all graphs are executed, changed and queried
+	const float EPSILON = 2 * 1E-6;  // Tolerance for GPU-CPU matrix multiplication differences
+	ASSERT_THAT(GRAPH_QUERIES, Lt(GRAPH_SIZE));
```
PiotrMrozikRobotec, line 149: This is nice :1st_place_medal:

test/src/asyncStressTest.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+	// Setup random graphs and helper structures
+	std::vector<RandomGraph> graphs;
+	std::vector<GraphRun> runs;
+	runs.resize(GRAPH_COUNT);
+	for (int g = 0; g < GRAPH_COUNT; ++g) {
+		graphs.emplace_back(GRAPH_SIZE);
+	}
+
+	for (int e = 0; e < GRAPH_EPOCHS; ++e) {
```
PiotrMrozikRobotec, line 167: We can think, if this EPOCHS logic, can be transferred into GTEST repeat mechanism, but not sure for now.

test/src/asyncStressTest.cpp
```diff
@@ -0,0 +1,242 @@
...
+				EXPECT_RGL_SUCCESS(rgl_graph_get_result_size(checkedNode, RGL_FIELD_XYZ_F32, &pointCount, &pointSize));
+				EXPECT_THAT(pointCount, Eq(run.expectedPointCount));
+
+				// Get results
+				std::vector<Vec3f> results(pointCount);
+				EXPECT_RGL_SUCCESS(rgl_graph_get_result_data(checkedNode, RGL_FIELD_XYZ_F32, results.data()));
+
+				// Verify results are correctly transformed
+				Vec3f expectedPoint = run.expectedTransform.at(checkedNode) * run.expectedPointWorld;
+				EXPECT_THAT(results.at(0).x(), FloatNear(expectedPoint.x(), EPSILON));
```
PiotrMrozikRobotec, line 222: I'm not sure about, however:
We can think about using only `EXPECT_EQ` in cases of equality.
In Your case, everything is alright, however for future uses in tests, we can eliminate stupid mistakes of wrongly used `EXPECT_THAT`.
According to documentation:
`WARNING: Equality matching via EXPECT_THAT(actual_value, expected_value) is supported, however, note that implicit conversions can cause surprising results. For example, EXPECT_THAT(some_bool, "some string") will compile and may pass unintentionally.`

test/src/asyncStressTest.cpp
```diff
@@ -0,0 +1,242 @@
...
+					EXPECT_THAT(results.at(i).y(), FloatEq(results.at(i - 1).y()));
+					EXPECT_THAT(results.at(i).z(), FloatEq(results.at(i - 1).z()));
+				}
+
+				// At some point, modify rays while the graph may be still running
+				if (checkedNode == run.updateRaysAfterCheckingNode) {
+					graph.setRandomRays();
+				}
+
+				run.nextNodeCheckOrderIndex += 1;
```
PiotrMrozikRobotec, line 238: Y not c++? :upside_down_face:

test/include/lidars.hpp
```diff
@@ -31,3 +31,24 @@ static std::vector<rgl_mat3x4f> makeLidar3dRays(float fov_x, float fov_y,
 
     return rays;
 }
+
+static std::vector<rgl_mat3x4f> makeGridOfParallelRays(Vec2f minCoord, Vec2f maxCoord, Vec2i count)
```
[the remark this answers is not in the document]
prybicki: Yes, I consider it very likely. A little library of basic building blocks to construct test- scenes, lidars or even graphs would be handy. 
Such methods could be also moved to src/ and be indirectly accessible through RGL (e.g. to build a uniform LiDAR firing pattern).

test/include/scenes.hpp
```diff
@@ -24,6 +24,13 @@ static rgl_entity_t makeEntity(rgl_mesh_t mesh= nullptr, rgl_scene_t scene=nullp
 	return entity;
 }
 
+static void setupSceneCube(Vec3f position)
+{
+	rgl_entity_t cube = makeEntity(makeCubeMesh());
+	rgl_mat3x4f pose = Mat3x4f::translation(position.x(), position.y(), position.z()).toRGL();
```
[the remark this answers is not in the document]
prybicki: A couple thoughts here:
- We have a bigger elephant in the room: `rgl_entity_set_pose`. Would you rename these too?
- [According to wikipedia](https://en.wikipedia.org/wiki/Pose_(computer_vision)), this is a fairly reasonable variable name:
  - _In the fields of [computing](https://en.wikipedia.org/wiki/Computing) and [computer vision](https://en.wikipedia.org/wiki/Computer_vision), pose (or spatial pose) represents the [position](https://en.wikipedia.org/wiki/Position_(geometry)) and [orientation](https://en.wikipedia.org/wiki/Orientation_(geometry)) of an object, usually in [three dimensions](https://en.wikipedia.org/wiki/Three_dimensions). Poses are often stored internally as [transformation matrices](https://en.wikipedia.org/wiki/Transformation_matrix).[[2]](https://en.wikipedia.org/wiki/Pose_(computer_vision)#cite_note-2)[[3]](https://en.wikipedia.org/wiki/Pose_(computer_vision)#cite_note-3) The term “pose” is largely synonymous with the term “transform”, but a transform may often include [scale](https://en.wikipedia.org/wiki/Scaling_(geometry)), whereas pose does not.[[4]](https://en.wikipedia.org/wiki/Pose_(computer_vision)#cite_note-4)[[5]](https://en.wikipedia.org/wiki/Pose_(computer_vision)#cite_note-5)_

test/src/asyncStressTest.cpp
```diff
@@ -0,0 +1,242 @@
...
+			);
+
+			rgl_node_t childNode = nullptr;
+			rgl_mat3x4f childTransformRGL = childTransform.toRGL();
+			EXPECT_RGL_SUCCESS(rgl_node_points_transform(&childNode, &childTransformRGL));
+			EXPECT_RGL_SUCCESS(rgl_graph_node_add_child(parentNode, childNode));
+			transformNodes.push_back(childNode);
+		}
+
+		setRandomRays();
```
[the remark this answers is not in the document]
prybicki: I would gladly put everything in the constructor, but this (re)randomizing rays is also done from the test, hence a separate function.  Added a comment. Changed the name to `generateRandomRays(),` which, IMHO, is more precise.

test/src/asyncStressTest.cpp
```diff
@@ -0,0 +1,242 @@
...
+	}
+
+	void setRandomRays()
+	{
+		std::uniform_real_distribution<float> randomFloat {-0.9, 0.9};
```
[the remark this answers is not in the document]
prybicki: You're right. I've considered a refactor but concluded that an explicit comment is enough here.

test/src/asyncStressTest.cpp
prybicki: This will be addressed by @nebraszka. She already did something like this in her PR.

test/src/asyncStressTest.cpp
prybicki: Agree. It is very tempting. Where GRAPH_EPOCHS would be defined? I guess it would be an external file (e.g. python script calling test binary).

test/src/asyncStressTest.cpp
prybicki: Great point. Replaced all `EXPECT_THAT` with `EXPECT_EQ` and `EXPECT_NEAR.`

test/src/asyncStressTest.cpp
prybicki: Why so?

test/src/asyncStressTest.cpp
prybicki: Jk jk, we remove logging from tests.

test/src/asyncStressTest.cpp
PiotrMrozik: I guess so.
In some future, we can prepare python script executing tests with proper flags like `--gtest_repeat`
Then all this could be programmed in some external init file for a python script. 
Fortunately we have bright tester ;)

test/src/asyncStressTest.cpp
PiotrMrozik: I just like ++ syntax.

test/src/asyncStressTest.cpp
```diff
@@ -0,0 +1,242 @@
...
+				}
+			}
+		};
+		dfs(parentTransformNode, Mat3x4f::identity());
+		return result;
+	}
+};
+
+/* Helper structure for data associated with a single graph and a single run-modify-check cycle */
+struct GraphRun {
```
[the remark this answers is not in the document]
PiotrMrozik: @prybicki ?

test/src/asyncStressTest.cpp
```diff
@@ -0,0 +1,247 @@
...
+		rayCount = { randomInt(randomGenerator), randomInt(randomGenerator) };
+		std::vector<rgl_mat3x4f> rays = makeGridOfParallelRays(raysXY, raysXY, {rayCount.x(), rayCount.y()});
+		EXPECT_RGL_SUCCESS(rgl_node_rays_from_mat3x4f(&useRaysNode, rays.data(), rays.size()));
+	}
+
+	std::unordered_map<rgl_node_t, Mat3x4f> getCumulativeTransform() const
+	{
+		std::unordered_map<rgl_node_t, Mat3x4f> result;
+		std::function<void(rgl_node_t, Mat3x4f)> dfs = [&](rgl_node_t currentBase, Mat3x4f parentCumulative) {
+			auto* currentTyped = dynamic_cast<TransformPointsNode*>(currentBase);
```
PiotrMrozik, line 105: `dynamic_pointer_cast` wouldn't be cleaner?

PiotrMrozik: Nah, I'm a stupid sandwich. NVM.

test/src/asyncStressTest.cpp
prybicki: I don't :)

test/src/asyncStressTest.cpp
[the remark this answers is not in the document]
prybicki: Did you mean `per test execution`? No test instancing in sight here.
Also, regarding using GTEST repeat mechanism - I'd vote not to use it for stress tests. Part of the stress is that there might be garbage (of whatever kind) from previous runs, and we want to catch such issues.

test/src/asyncStressTest.cpp
prybicki: After giving it more thought - I'd vote not to use it for stress tests. Part of the stress is that there might be garbage (of whatever kind) from previous runs, and we want to catch such issues.

test/include/scenes.hpp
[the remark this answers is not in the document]
prybicki: Whatever we choose, I think that's a task extending beyond the scope of this PR.

test/src/asyncStressTest.cpp
PiotrMrozik: That's the good point.

test/include/scenes.hpp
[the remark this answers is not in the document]
PiotrMrozik: We can leave it for now.

PiotrMrozik (approved):

PR #89, opened 2023-01-24
ROS2 standalone build - setup.py review suggestions

prybicki:
Changelog:
- Changes tabs to spaces (to conform PEPs)
- Change file layout, wrap global variables into a class, so that helper functions can be moved to the bottom
- Fixes not checking the CUDA patch version, which is significant on Windows
- Replace error-prone string joining with appending to list.
- Introduce `run_subprocess_command` and run_system_command throwing an exception on failure.
- Change some args defaults from `None` to "" where appropriate.
- Build with all cores by default

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,17 +7,141 @@ import subprocess
 import shutil
 import argparse
 
-# Global variables (default values for Linux)
-CUDA_MIN_VER_MAJOR = 11
-CUDA_MIN_VER_MINOR = 2
-CMAKE_GENERATOR = "'Unix Makefiles'"
-VCPKG_INSTALL_DIR = os.path.join("external", "vcpkgA")
-VCPKG_TAG = "2022.08.15"
-VCPKG_EXEC = "vcpkg"
-VCPKG_BOOTSTRAP = "bootstrap-vcpkg.sh"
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -7,148 +7,171 @@
...
+                        help="Remove build directory before cmake")
+    parser.add_argument("--with-ros2", action='store_true',
+                        help="Build RGL with ROS2 extension")
+    parser.add_argument("--with-ros2-standalone", action='store_true',
+                        help="Build RGL with ROS2 extension in standalone mode")
+    parser.add_argument("--cmake", type=str, nargs=1, default="",
+                        help="Pass arguments to cmake. Usage: --cmake=\"args...\"")
+    if on_linux():
+        parser.add_argument("--make", type=str, nargs=1, default=f"-j{os.cpu_count()}", dest="build_args",
+                            help="Pass arguments to make. Usage: --make=\"args...\"")
```
msz-rai, line 55: It would be nice to inform the user about the default value.

msz-rai (approved):
Great refactor! Thank you.
I have found that `--cmake`, `--make`, and `--ninja` arguments are not working. Please address the comments.

PR #93, opened 2023-02-07
Add time to scene

msz-rai:


src/Time.hpp
```diff
@@ -0,0 +1,53 @@
...
+{
+	using SecNanosecPair = std::pair<int32_t, uint32_t>;
+
+	static Time zero()
+	{ return Time(0, 0); }
+
+	Time(int32_t seconds, uint32_t nanoseconds)
+	{ time = std::make_pair(seconds, nanoseconds); }
+	Time(double seconds)
+	{ time = std::make_pair(static_cast<int32_t>(seconds), static_cast<uint32_t>(seconds * 1.0e9)); };
```
PiotrMrozikRobotec, line 31: ditto

src/Time.hpp
```diff
@@ -0,0 +1,53 @@
...
+	{
+		auto msg = builtin_interfaces::msg::Time();
+		msg.sec = time.first;
+		msg.nanosec = time.second;
+		return msg;
+	}
+	#endif
+
+	inline bool operator==(const Time& other) const
+	{ return time.first == other.time.first && time.second == other.time.second; }
```
PiotrMrozikRobotec, line 49: I don't think this is necessary. from ccp14 Yu can easily compare pairs of trivial types.

PiotrMrozikRobotec (commented):
The big question is the legitimacy of two separate subfields for seconds and nanoseconds. 
The real purpose of that defines if that should be rewritten or not. 
I suggest using one double and formatting it only if needed when the value is exported to ros.

src/Time.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+#ifdef RGL_BUILD_ROS2_EXTENSION
+#include <builtin_interfaces/msg/time.hpp>
+#endif
+
+struct Time
+{
+	static Time zero() { return Time(0); }
+	static Time seconds(double seconds) { return Time(static_cast<uint64_t>(seconds * 1.0e9)); }
+	static Time nanoseconds(uint64_t nanoseconds) { return Time(nanoseconds); }
```
PiotrMrozikRobotec, line 25: Can You explain why You decided to make private constructor and overlay it with few public static methods?
This is not a mistake. Just want to fully understand intention.

PiotrMrozikRobotec (approved):
I have one question: not even a mistake. Approved anyway.

src/Time.hpp
msz-rai: I wanted to make the code cleaner when creating Time objects. Having constructors with parameters that differ on param types, there is no explicit information on what units you pass.
For example, `Time(1)` and `Time(1.0f)` will make a huge difference in what you get, and you need to check the implementation of every constructor to know the value of the time you have created.

PR #94, opened 2023-02-07
Remove ROS2 executor from Ros2PublishPointsNode

msz-rai:
An executor is not needed since we only publish data.


PiotrMrozikRobotec (approved):
You indeed removed ROS2 executor from publish points node

PR #96, opened 2023-02-10
Convert node write_pcd_file to API call

msz-rai:


include/rgl/api/extensions/pcl.h
```diff
@@ -16,6 +16,17 @@
 
 #include <rgl/api/core.h>
 
+/******************************** GRAPH ********************************/
+
+/**
+ * Obtains the point cloud of any node in the graph and saves to given file.
+ * Output file will be saved in PCD format (https://pointclouds.org/documentation/tutorials/pcd_file_format.html)
+ * @param node Node to get point cloud from.
+ * @param file_path Path to the output pcd file.
```
PiotrMrozikRobotec, line 25: There are any requrements? Absolute? Relative?
That should be in param descriprion.

src/api/apiPcl.cpp
```diff
@@ -12,54 +12,81 @@
...
-		RGL_API_LOG("rgl_node_points_downsample(node={}, leaf=({}, {}, {}))", repr(node), leaf_size_x, leaf_size_y, leaf_size_z);
+		RGL_API_LOG("rgl_graph_write_pcd_file(node={}, file={})", repr(node), file_path);
+		CHECK_ARG(file_path != nullptr);
+		CHECK_ARG(file_path[0] != '\0');
 
-		createOrUpdateNode<DownSamplePointsNode>(node, Vec3f{leaf_size_x, leaf_size_y, leaf_size_z});
+		auto pointCloudNode = Node::validatePtr<IPointsNode>(node);
+
+		if (!pointCloudNode->hasField(XYZ_F32)) {
+			throw InvalidAPIObject(fmt::format("Saving PCD file {} failed - requested node hasn't field XYZ.", file_path));
```
PiotrMrozikRobotec, line 42: `node hasn't field`
Same case as in the last PR. Correct but misleading use of language.

PiotrMrozikRobotec (changes requested):

include/rgl/api/extensions/pcl.h
msz-rai: There is no restriction. It could be absolute or relative.

PiotrMrozikRobotec (approved):

PR #99, opened 2023-02-17
Add api surface tests

nebraszka:


test/include/utils.hpp
```diff
@@ -22,18 +20,35 @@
         EXPECT_EQ(actual, expected);                                    \
         const char* error_string;                                       \
         rgl_get_last_error_string(&error_string);                       \
-        EXPECT_THAT(error_string, HasSubstr(error_prefix));       \
-        EXPECT_THAT(error_string, HasSubstr(error_detail));         \
+        EXPECT_THAT(error_string, HasSubstr(error_prefix));       		\
+        EXPECT_THAT(error_string, HasSubstr(error_detail));         	\
```
prybicki, line 24: The purposes of `error_prefix` and `error_detail` seem the same, yet we have two different names.
An easy fix would be to rename both to `error_substring1`, `error_substring2`.
A more ambitious way would be to replace both with `__VA_ARGS__`.
Hint: something like this might work:
```cpp
EXPECT_RGL_STATUS(actual, expected, ...)
```
and then:
```cpp
for (auto&& substr : { __VA_ARGS__ }) { EXPECT_THAT(error_string, HasSubstr(substr)); }
```

test/src/apiSurfaceTests.cpp
```diff
@@ -118,11 +120,11 @@ TEST_F(APISurfaceTests, rgl_entity_create_destroy)
...
-	EXPECT_RGL_INVALID_OBJECT(rgl_entity_destroy(entity), "Object does not exist: Entity");
+	EXPECT_RGL_OBJECT_NOT_EXISTS(rgl_entity_destroy(entity), "Object does not exist: Entity");
 
 	// Invalid destroy
 	EXPECT_RGL_INVALID_ARGUMENT(rgl_entity_destroy(nullptr), "entity != nullptr");
-	EXPECT_RGL_INVALID_OBJECT(rgl_entity_destroy((rgl_entity_t) 0x1234), "Object does not exist: Entity 0x1234");
+	EXPECT_RGL_OBJECT_NOT_EXISTS(rgl_entity_destroy((rgl_entity_t) 0x1234), "Object does not exist: Entity 0x1234");
 }
 
 TEST_F(APISurfaceTests, rgl_entity_set_pose)
```
PiotrMrozikRobotec, line 130: Not to Ur PR, but:
`rgl_entity_set_pose` Is terribly wrong named.
Pose in 3d envs, and CG means something very specific, and as far as I know, we are using skinned meshes, which can contain poses.

Should be renamed to FE:
`rgl_entity_set_local_transform`

PiotrMrozikRobotec (changes requested):
Almost all the tests in 
`test/src/apiSurfaceTests.cpp`
Should be rewritten into TEST_P in order to fully cover all permutations of parameters. 
Now we are testing some subset of cases.
After that, we can continue corner cases/logic of tests review. 

Also remember the "use local variables/structs once" rule.

test/include/utils.hpp
```diff
@@ -22,18 +20,35 @@
...
-#define EXPECT_RGL_INVALID_OBJECT(status, type) EXPECT_RGL_STATUS(status, RGL_INVALID_API_OBJECT, "Object does not exist", type)
+#define EXPECT_RGL_OBJECT_NOT_EXISTS(status, type) EXPECT_RGL_STATUS(status, RGL_INVALID_API_OBJECT, "Object does not exist", type)
+#define EXPECT_RGL_NODE_TYPE_MISMATCH(status, type) EXPECT_RGL_STATUS(status, RGL_INVALID_API_OBJECT, "Node type mismatch", type)
 #define EXPECT_RGL_INVALID_ARGUMENT(status, error) EXPECT_RGL_STATUS(status, RGL_INVALID_ARGUMENT, "Invalid argument", error)
+#define EXPECT_RGL_INVALID_PIPELINE(status, error) EXPECT_RGL_STATUS(status, RGL_INVALID_PIPELINE, "attempted to remove child", error)
 
+#define FILENAME (strrchr(__FILE__, std::filesystem::path::preferred_separator) ? strrchr(__FILE__, std::filesystem::path::preferred_separator) + 1 : __FILE__)
 
-struct RGLAutoCleanupTest : public ::testing::Test {
+struct RGLAutoSetUp : public ::testing::Test {
```
msz-rai, line 35: I'd leave `Test` postfix: `RGLAutoSetUpTest`.

test/src/graphTest.cpp
```diff
@@ -6,7 +6,11 @@
 
 #include <math/Mat3x4f.hpp>
 
-class Graph : public RGLAutoCleanupTest {};
+class Graph : public RGLAutoSetUp
```
msz-rai, line 9: Please rename this class (e.g., `GraphCase`). Graph already exists in our repo.

test/src/loggingTests.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+	// Expected log levels should be written in the file
+	std::string logFile = readFileStr(logFilePath);
+	EXPECT_THAT(logFile, Not(HasSubstr("[trace]: This is RGL trace log.")));
+	EXPECT_THAT(logFile, HasSubstr("[info]: This is RGL info log."));
+	EXPECT_THAT(logFile, HasSubstr("[warning]: This is RGL warn log."));
+	EXPECT_THAT(logFile, HasSubstr("[error]: This is RGL error log."));
+	EXPECT_THAT(logFile, HasSubstr("[critical]: This is RGL critical log."));
+	EXPECT_RGL_SUCCESS(rgl_configure_logging(RGL_LOG_LEVEL_OFF, nullptr, false));
+}
```
msz-rai, line 41: Logging test with Tape is missing. Deleted [here](https://github.com/RobotecAI/RobotecGPULidar/commit/ad935fc3d79f4730dc4358397402cf24e44d1f1c#diff-0d7f34bfea4a4a23209b9755b3da43cbee7825a3635301c932dadd23641c3841L32)

test/src/tapeSurfaceTest.cpp
```diff
@@ -13,7 +13,10 @@
 
 #include <math/Mat3x4f.hpp>
 
-class Tape : public RGLAutoCleanupTest {};
+class Tape : public RGLAutoSetUp
```
msz-rai, line 16: I'd rename this class - potential collision in the future (same as Graph).

msz-rai (changes requested):

test/src/apiSurfaceTests.cpp
```diff
@@ -131,142 +133,415 @@ TEST_F(APISurfaceTests, rgl_entity_set_pose)
 
 	// Invalid args
 	EXPECT_RGL_INVALID_ARGUMENT(rgl_entity_set_pose(nullptr, nullptr), "entity != nullptr");
-	EXPECT_RGL_INVALID_ARGUMENT(rgl_entity_set_pose(entity,  nullptr), "local_to_world_tf != nullptr");
-	EXPECT_RGL_INVALID_OBJECT(rgl_entity_set_pose((rgl_entity_t) 0x1234, &identity), "Object does not exist: Entity 0x1234");
+	EXPECT_RGL_INVALID_ARGUMENT(rgl_entity_set_pose(entity,  nullptr), "transform != nullptr");
```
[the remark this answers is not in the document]
prybicki: That could cause weird silent bugs.

docs/PclExtension.md
```diff
@@ -0,0 +1,33 @@
...
+
+- Downsampling
+- Writing to PCD file
+- Visualization
+
+For operations listed above, the extension uses [Point Cloud Library](https://pointclouds.org/).
+
+## Building
+
+Before building RGL PCL extension, it is necessary to install the required dependencies. Run the setup script with `--install-pcl-deps` flag to download and install them. It could take some time (Point Cloud Library is built from source):
```
prybicki, line 13:
```suggestion
After building RGL PCL extension, it is necessary to install the required dependencies. Run the setup script with `--install-pcl-deps` flag to download and install them. It could take some time (Point Cloud Library is built from source):
```

PR #103, opened 2023-02-28
Make TemporalMergePointsNode use Host memory (with tests)

prybicki:


msz-rai (approved):
LGTM :+1: pros for the test.
Shouldn't SpatialMergePointsNode be host-based by default also?

PR #104, opened 2023-03-02
[v13] Feature/use points from array deep test

ghost:


test/include/utils.hpp
```diff
@@ -22,13 +22,14 @@
         EXPECT_EQ(actual, expected);                                    \
         const char* error_string;                                       \
         rgl_get_last_error_string(&error_string);                       \
-        EXPECT_THAT(error_string, HasSubstr(error_prefix));       \
-        EXPECT_THAT(error_string, HasSubstr(error_detail));         \
+        EXPECT_THAT(error_string, testing::HasSubstr(error_prefix));       \
```
nebraszka, line 25: See @prybicki  [comment](
https://github.com/RobotecAI/RobotecGPULidar/pull/99#discussion_r1111896210):
> The purposes of `error_prefix` and `error_detail` seem the same, yet we have two different names. An easy fix would be to rename both to `error_substring1`, `error_substring2`. A more ambitious way would be to replace both with `__VA_ARGS__`. Hint: something like this might work:
> 
> ```c++
> EXPECT_RGL_STATUS(actual, expected, ...)
> ```
> 
> and then:
> 
> ```c++
> for (auto&& substr : { __VA_ARGS__ }) { EXPECT_THAT(error_string, HasSubstr(substr)); }
> ```



> The purposes of `error_prefix` and `error_detail` seem the same, yet we have two different names. An easy fix would be to rename both to `error_substring1`, `error_substring2`. A more ambitious way would be to replace both with `__VA_ARGS__`. Hint: something like this might work:
> 
> ```c++
> EXPECT_RGL_STATUS(actual, expected, ...)
> ```
> 
> and then:
> 
> ```c++
> for (auto&& substr : { __VA_ARGS__ }) { EXPECT_THAT(error_string, HasSubstr(substr)); }
> ```

test/src/graph/nodes/FromArrayPointsNodeTest.cpp
```diff
@@ -0,0 +1,102 @@
...
+    auto inPoints = GeneratePointsArray(pointsCount);
+
+    rgl_node_t usePointsNode = nullptr;
+    rgl_node_t formatNode = nullptr;
+
+    EXPECT_RGL_SUCCESS(rgl_node_points_from_array(&usePointsNode, inPoints.data(), inPoints.size(), pointFields.data(), pointFields.size()));
+    ASSERT_THAT(usePointsNode, testing::NotNull());
+
+    EXPECT_RGL_SUCCESS(rgl_node_points_format(&formatNode, pointFields.data(), pointFields.size()));
+    ASSERT_THAT(formatNode, testing::NotNull());
```
prybicki, line 75: This use_case seems more like a testing format than fromArray.

test/src/graph/nodes/FromArrayPointsNodeTest.cpp
```diff
@@ -0,0 +1,116 @@
...
+        expectedIntensity[i] = expectedPoints[i].intensity;
+    }
+
+    for (auto field : pointFields) {
+        int32_t outCount, outSizeOf;
+        EXPECT_RGL_SUCCESS(rgl_graph_get_result_size(usePointsNode, field, &outCount, &outSizeOf));
+        EXPECT_EQ(outCount, inPoints.size());
+        EXPECT_EQ(outSizeOf, getFieldSize(field));
+
+        void* outData = malloc(outCount * outSizeOf);
```
msz-rai, line 98: Use `std::vector<Field<field>::type> outData` This will free data automatically, and you will avoid typecasting.

test/src/graph/nodes/FromArrayPointsNodeTest.cpp
```diff
@@ -0,0 +1,116 @@
...
+        EXPECT_RGL_SUCCESS(rgl_graph_get_result_data(usePointsNode, field, outData));
+
+        for (int i = 0; i < outCount; ++i) {
+            if (field == XYZ_F32) {
```
msz-rai, line 102: What about `switch`?

msz-rai (changes requested):
Solid test design. Please address the comments.

msz-rai (approved):
LGTM. Please do not get baited on overengineering next time :stuck_out_tongue:

PR #107, opened 2023-03-06
Release v12

msz-rai:


include/rgl/api/core.h
```diff
@@ -308,6 +328,18 @@ rgl_entity_destroy(rgl_entity_t entity);
 RGL_API rgl_status_t
 rgl_entity_set_pose(rgl_entity_t entity, const rgl_mat3x4f *local_to_world_tf);
 
+/******************************** SCENE ********************************/
+
+/**
+ * Sets time to the given scene.
+ * Time indicates a specific point in time, relative to a clock's 0 point.
```
prybicki, line 335: What does `clock's 0 point` mean? What clock?

include/rgl/api/core.h
```diff
@@ -308,6 +328,18 @@ rgl_entity_destroy(rgl_entity_t entity);
...
+ * Timestamp is used to fill field RGL_FIELD_TIME_STAMP_F64 or for ROS2 publishing.
+ * @param scene Scene where time will be set. Pass NULL to use the default scene.
+ * @param nanoseconds Timestamp in nanoseconds.
+ */
+RGL_API rgl_status_t
+rgl_scene_set_time(rgl_scene_t scene, uint64_t nanoseconds);
```
prybicki, line 341: I wonder if it is the right interface. Although it will work for the next 500 years, I doubt it is natural to express time in nanoseconds. Could you have a look what interface would be most convenient for currently existing plugins and share thoughts?

prybicki (changes requested):

include/rgl/api/core.h
msz-rai:
```suggestion
 * Time indicates a specific point in the simulation timeline when raytrace is performed.
```
Is that cleaner?

include/rgl/api/core.h
msz-rai: I have modeled it on the ROS2 time format: two ints32 for seconds and nanoseconds. I have simplified that into one field. My wonder was to use `double seconds` or `uint64_t nanoseconds`. The first option was more natural, but no resolution was guaranteed (milliseconds or nanoseconds?). Since this time is used to work with ROS2, I wanted to ensure that the time set will be exactly the same as provided.

README.md
```diff
@@ -70,32 +75,29 @@ Build instructions:
...
+   - See `./setup.py --help` for usage information.
 
 ## Building on Windows
 
-1. Install [CUDA Toolkit](https://developer.nvidia.com/cuda-downloads) **11.4.4+**.
-2. Download [NVidia OptiX](https://developer.nvidia.com/designworks/optix/downloads/legacy) **7.2**.
+1. Install [Microsoft Visual Studio](https://visualstudio.microsoft.com/pl/downloads/) (Visual Studio 2019 when using ROS2 extension) with **C++ CMake tools for Windows** component.
+2. Install [CUDA Toolkit](https://developer.nvidia.com/cuda-downloads) **11.4.4+**.
+3. Download [NVidia OptiX](https://developer.nvidia.com/designworks/optix/downloads/legacy) **7.2**.
    - use the default location or set environment variable `OptiX_INSTALL_DIR`
```
msz-rai, line 91:
```suggestion
   - install the framework and set environment variable `OptiX_INSTALL_DIR`
```
Environment variable is not optional since we are checking it in the `setup.py` script.

nebraszka (approved):

PR #110, opened 2023-03-13
Transform points node deep test

PiotrMrozik:


test/include/utils.hpp
```diff
@@ -131,3 +143,15 @@ static rgl_mesh_t loadMesh(std::filesystem::path path)
...
+{
+        std::vector<TestPointStruct> points;
+        for (int i = 0; i < count; ++i) {
+            auto currentPoint = TestPointStruct { .xyz = { i, i + 1, i + 2 }, .isHit = i % 2, .intensity = 100 };
+            currentPoint.xyz = Mat3x4f::fromRGL(transform) * currentPoint.xyz;
+            points.emplace_back(currentPoint);
+
+        }
+        return points;
+}
```
msz-rai, line 157: It could be moved to `RGLGraphTest`/`RGLPointFieldsTest` struct when you will apply the previous comment.

test/src/graphTest.cpp
msz-rai, line 1: Why?

msz-rai (approved):
LGTM. Please consider suggestions in the comments.

test/include/utils.hpp
```diff
@@ -38,12 +42,22 @@ struct RGLAutoCleanupTest : public ::testing::Test {
...
+        IS_HIT_I32,
+        INTENSITY_F32
+    };
+};
+
+struct TestPointStruct {
+    Field<XYZ_F32>::type xyz;
+    Field<IS_HIT_I32>::type isHit;
+    Field<INTENSITY_F32>::type intensity;
 };
```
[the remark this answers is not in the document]
PiotrMrozik: Ad 1.
I see no problem with aggregating those two structs.

Ad 2.
I had a problem with naming it. 
I see Your concern, however name `RGLPointFieldsTest` is also far from perfect...

test/src/graphTest.cpp
PiotrMrozik: I messed up one thing during conflict resolving.
Somewhere between I clang format it, and I don't want to waste time reverting it.
Just ignore it.

test/src/graph/nodes/TransformPointsNodeTest.cpp
```diff
@@ -0,0 +1,89 @@
...
+#include <utils.hpp>
+
+class TransformPointsNodeTest : public RGLAutoCleanupTestWithParam<std::tuple<int, rgl_mat3x4f>>, public RGLPointTestHelper {
+};
+
+INSTANTIATE_TEST_SUITE_P(
+    TransformPointsNodeTests, TransformPointsNodeTest,
+    testing::Combine(
+        testing::Values(1, 10, 100000),
+        testing::Values(identityTestTransform, translationTestTransform, rotationTestTransform, scalingTestTransform, complexTestTransform)));
```
prybicki, line 14: 300 ms total. How about `1, 100, 10000`?

test/src/graph/nodes/TransformPointsNodeTest.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+TEST_P(TransformPointsNodeTest, invalid_arguments)
+{
+    auto transform = std::get<1>(GetParam());
```
prybicki, line 18: Magic constant. IMHO, this could be easily eliminated by adding named indices (e.g. constexprs) together with `TransformPointsNodeTest` definition, and then you'd have:
```suggestion
    auto transform = std::get<TEST_PARAM_TRANSFORM_ID>(GetParam());
```

prybicki: Also, this `std::get<1>(GetParam())` is not necessary, because later lambda is called.

test/src/graph/nodes/TransformPointsNodeTest.cpp
PiotrMrozik: I am fine with everything greater than the number of CU in a typical user GPU.

prybicki: Good point. Let's define this value. At the moment, this would be 20k.

test/src/graph/nodes/TransformPointsNodeTest.cpp
```diff
@@ -0,0 +1,89 @@
...
+TEST_P(TransformPointsNodeTest, use_case)
+{
+    auto [pointsCount, transform] = GetParam();
+    auto inPoints = GenerateTestPointsArray(pointsCount);
+
+    rgl_node_t usePointsNode = nullptr;
+    rgl_node_t transformNode = nullptr;
+
+    EXPECT_RGL_SUCCESS(rgl_node_points_from_array(&usePointsNode, inPoints.data(), inPoints.size(), pointFields.data(), pointFields.size()));
+    ASSERT_THAT(usePointsNode, testing::NotNull());
```
prybicki, line 57: Consider making `UsePointsNode` a part of `RGLPointTestHelper`.

test/src/graph/nodes/TransformPointsNodeTest.cpp
```diff
@@ -0,0 +1,89 @@
...
+    EXPECT_RGL_SUCCESS(rgl_graph_run(usePointsNode));
+
+    int32_t outCount, outSizeOf;
+    EXPECT_RGL_SUCCESS(rgl_graph_get_result_size(transformNode, XYZ_F32, &outCount, &outSizeOf));
+
+    EXPECT_EQ(outCount, inPoints.size());
+    EXPECT_EQ(outSizeOf, getFieldSize(XYZ_F32));
+
+    std::vector<::Field<XYZ_F32>::type> outData;
+    outData.reserve(outCount);
```
prybicki, line 73: `reserve` does not change size!
```suggestion
    outData.resize(outCount);
```

test/src/graph/nodes/TransformPointsNodeTest.cpp
```diff
@@ -0,0 +1,89 @@
...
+    // TODO When we are testing big values of point translation, numerical errors appears.
+    //  For example for 100000 unit of translation, error after rotation can extend 0.001 unit.
+    //  To investigate in better times.
+
+    // TODO 2 In the future, this and other tests, should check if every possible field has proper value, even if intact.
+    for (int i = 0; i < pointsCount; ++i) {
+        // For now use percent (0.1%) of value as error treshold.
+        EXPECT_NEAR(expectedPoints[i].xyz[0], outData[i][0], (outData[i][0]/ 1000));
+        EXPECT_NEAR(expectedPoints[i].xyz[1], outData[i][1], (outData[i][1] / 1000));
+        EXPECT_NEAR(expectedPoints[i].xyz[2], outData[i][2], (outData[i][2] / 1000));
```
prybicki, line 87: You'd get out-of-bounds error if you'd use `at()` here (which I recommend).

prybicki (changes requested):
One important issue to address (reserve) and a few suggestions.

test/src/graph/nodes/TransformPointsNodeTest.cpp
PiotrMrozik: Let's define it as 20k for now in some test utils const.

test/src/graph/nodes/TransformPointsNodeTest.cpp
PiotrMrozik: Let's stick to:
` auto [_, transform] = GetParam();`

prybicki (approved):

PR #116, opened 2023-03-20
Streams synchronization

PiotrMrozik:


test/CMakeLists.txt
```diff
@@ -13,7 +13,8 @@ set(RGL_TEST_FILES
     src/testMat3x4f.cpp
     src/graph/nodes/FromArrayPointsNodeTest.cpp
     src/ArrayTest.cpp
-)
+    src/streams/streamsTest.cpp
+    src/streams/testKernel.cu)
```
prybicki, line 17: I think `src/synchronization` is a slightly better and future-proof name.

test/src/streams/streamsTest.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include <utils.hpp>
+#include "testKernel.hpp"
+
+using namespace testing;
+
+class StreamsTest : public RGLAutoCleanupTest {
+protected:
+};
+
+TEST_F(StreamsTest, basic_streams)
```
prybicki, line 12:
```suggestion
TEST_F(SyncTests, GraphAndCopyStream)
```

test/src/streams/streamsTest.cpp
```diff
@@ -0,0 +1,68 @@
...
+{
+    // Prepare two non-default streams.
+    cudaStream_t graphStream;
+    cudaStream_t copyStream;
+
+    cudaStreamCreate(&graphStream);
+    cudaStreamCreate(&copyStream);
```
prybicki, line 19: Check CUDA errors everywhere.

test/src/streams/streamsTest.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+    int N = 20000;
+    size_t size = N * sizeof(float);
+
+    // Allocate input vector h_A in host memory
+    float* h_A = (float*)malloc(size);
```
prybicki, line 25: Though it's just a test, variable names could be a bit more readable.

test/src/streams/streamsTest.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+    // Initialize input vectors
+    for (int i = 0; i < N; ++i) {
+        h_A[i] = i;
+    }
+    // Create event for sensitive job (kernel) end.
+    cudaEvent_t jobDoneEvent;
+    // Create event without timer (used for profiling).
+    // This way, we have got some speed, and also according to CUDA documentation, this flag is neccessary to make sure that multi-stream synchronization is bulltetproof.
```
prybicki, line 34: Can you link the doc regarding the "bulletproofness" of not-timing events? Does timing event fail to synchronize? This is important info.

test/src/streams/streamsTest.cpp
prybicki: You're using pageable memory here. In real world scenario, pinned memory will be used (`cudaHostAlloc`/`cudaMallocHost`). Please make sure it works well. Also, it would be a good idea to inspect this test case under NVIDIA Nsight Systems. You should have one installed together with CUDA Toolkit.
Example screenshot:
image

test/src/streams/streamsTest.cpp
```diff
@@ -0,0 +1,68 @@
...
+        EXPECT_EQ(temp, i * i);
+    }
+
+    cudaStreamDestroy(graphStream);
+    cudaStreamDestroy(copyStream);
+
+    cudaEventDestroy(jobDoneEvent);
+
+    free(h_A);
+    cudaFree(d_A);
```
prybicki, line 67: It's better (and often required) to use `cudaFreeAsync` in pair with `cudaMallocAsync`.

test/src/streams/streamsTest.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+    // Run sensitive job that we forecast, it takes some time.
+    testKernelWrapper(N, d_A, graphStream);
+
+    // Record synchronization event to the stream after sensitive job.
+    cudaEventRecord(jobDoneEvent, graphStream);
+
+    // Synchronize streams. This is pass or not for the test.
+    //  All API Calls in given stream, will wait for the event  to be completed.
+    cudaStreamWaitEvent(copyStream, jobDoneEvent);
```
prybicki, line 50: Here, I'd also consider adding some more work to the graph stream after the event (e.g. dummy `cudaMallocAsync`, `cudaMemsetAsync`, `cudaFreeAsync`. It's hard to do any additional checks, but it makes graph stream load more realistic (simulating other nodes). Also, it will allow us to make sure (using NVIDIA Nsight) that the work in the graph streams continues when copy stream works.

test/src/streams/testKernel.cu
```diff
@@ -0,0 +1,31 @@
...
+    } while (false)
+
+__global__ void testKernel(size_t pointCount, float* infloats)
+{
+    // Waste some time
+    clock_value_t start = clock64();
+    clock_value_t cycles_elapsed;
+    do {
+        cycles_elapsed = clock64() - start;
+    } while (cycles_elapsed < 2000000000);
```
prybicki, line 19: Is it O(second)?

test/CMakeLists.txt
```diff
@@ -43,6 +44,8 @@ target_include_directories(RobotecGPULidar_test PRIVATE
     ${CMAKE_CURRENT_SOURCE_DIR}/../src
     ${CMAKE_CURRENT_SOURCE_DIR}/../include
     ${CMAKE_CURRENT_SOURCE_DIR}/include
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/synchronization
+
```
PiotrMrozik:
```suggestion
```

PR #117, opened 2023-03-21
Replace validate() with onInputChange()

prybicki:


src/graph/SpatialMergePointsNode.cpp
```diff
@@ -31,12 +31,13 @@ void SpatialMergePointsNode::setParameters(const std::vector<rgl_field_t>& field
...
 
-void SpatialMergePointsNode::validate()
+void SpatialMergePointsNode::onInputChange()
 {
 	pointInputs = Node::filter<IPointsNode>(this->inputs);
 
-	if (pointInputs.empty()) {
-		auto msg = "expected at least one IPointsNode input for SpatialMergePointsNode";
+	if (pointInputs.size() < 2) {
+		// Having just one input does not make sense, because it is equivalent to removing this Node.
```
PiotrMrozik, line 39: I would stop right there for a while.
In the case of single input, this node is just a pass-through. This is not a mistake, just useless node.
I can imagine situation when we create a graph with spatial merge inside, only one input. This graph is working for some time, and when the situation is ready, we add second input (node or subgraph) to an already created and working graph.
This is not a no-go, however, re-think it.

PiotrMrozik (approved):
I like it.
Only few small questions.

src/graph/SpatialMergePointsNode.cpp
prybicki: So, why not add SpatialMerge when it's actually needed? Having it operate on just 1 input smells very much like a bug  / unintended situation.

prybicki: So, why not add SpatialMerge when it's actually needed? Having it operate on just 1 input smells very much like a bug  / unintended situation.

src/graph/SpatialMergePointsNode.cpp
```diff
@@ -48,10 +49,9 @@ void SpatialMergePointsNode::validate()
 		}
 
 		// Check input pointcloud has required fields
-		for (const auto& requiredField : std::views::keys(mergedData)) {
+		for (const auto& requiredField : getRequiredFieldList()) {
```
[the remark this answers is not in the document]
prybicki: SSOT

src/macros/cuda.hpp
```diff
@@ -36,6 +36,16 @@ do
 }                                                                                                   \
 while(false)
 
+#define CHECK_CUDA_NO_THROW(call) \
```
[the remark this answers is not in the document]
prybicki: I have been bitten once again by throwing destructor and this time I didn't tolerate it.
I will ask @nebraszka to inspect all other destructors, good task for a fresh Junior :)

src/graph/Interfaces.hpp
```diff
@@ -78,33 +86,44 @@ struct IPointsNode
...
+	virtual VArray::ConstPtr getFieldData(rgl_field_t field, cudaStream_t stream) const override
 	{ return input->getFieldData(field, stream); }
 
 protected:
-	IPointsNode::Ptr input;
+	IPointsNode::Ptr input {0};
 };
 
-struct IPointsNodeMultiInput : IPointsNode
+struct INoInputNode : virtual Node
```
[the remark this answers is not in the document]
prybicki: Hahaha, yeah, that's funny.

src/graph/SpatialMergePointsNode.cpp
PiotrMrozik: I partially agree with You.
I would warn the user about it, but still let it.

prybicki: I get your reasoning; also, consider the fact that the warning would land in log file and / or stdout, which requires additional effort to see in a typical RGL usage scenario (engine plugins). Therefore, I prefer an error, which is a minor inconvenience if the passthrough was desired and a huge helper if it was not.

PR #119, opened 2023-03-22
Fix clang-format

nebraszka:


prybicki (commented):
Looks very good; solid work!
Leaving a few improvement suggestions to try out.
Also, please add instructions to install the hook locally (a section in README or a separate file, leaving the decision which is best for you).

PR #120, opened 2023-03-23
Add with-all flag to setup.

PiotrMrozik:


CMakeLists.txt
```diff
@@ -26,6 +26,8 @@ set(RGL_BUILD_TESTS ON CACHE BOOL
 set(RGL_BUILD_TOOLS ON CACHE BOOL "Enables building RGL executable tools")
 
 # Extensions configuration
+set(RGL_BUILD_ALL OFF CACHE BOOL
+        "Enables all possible extensions, tools and tests.")
```
msz-rai, line 30: I'd prefer not to pollute CMakeLists with the flag `all`. We can manage it in the `setup.py` script.

setup.py
```diff
@@ -118,6 +120,7 @@ def is_cuda_version_ok():
     # Build
     cmake_args = [
         f"-DCMAKE_TOOLCHAIN_FILE={os.path.join(cfg.VCPKG_INSTALL_DIR, 'scripts', 'buildsystems', 'vcpkg.cmake') if args.with_pcl else ''}",
+        f"-DRGL_BUILD_ALL={'ON' if args.with_all else 'OFF'}",
         f"-DRGL_BUILD_PCL_EXTENSION={'ON' if args.with_pcl else 'OFF'}",
         f"-DRGL_BUILD_ROS2_EXTENSION={'ON' if args.with_ros2 else 'OFF'}",
         f"-DRGL_BUILD_ROS2_EXTENSION_STANDALONE={'ON' if args.with_ros2_standalone else 'OFF'}"
```
msz-rai, line 126:
```suggestion
        f"-DRGL_BUILD_PCL_EXTENSION={'ON' if args.with_pcl or args.with_all else 'OFF'}",
        f"-DRGL_BUILD_ROS2_EXTENSION={'ON' if args.with_ros2 or args.with_all else 'OFF'}",
        f"-DRGL_BUILD_ROS2_EXTENSION_STANDALONE={'ON' if args.with_ros2_standalone or args.with_all else 'OFF'}"
```
Suggestion to the previous comment.

msz-rai: I have realized that there is no flag to build tests and tools, but I see no obstacles to adding them (`--with-tools`, `--with-tests`). What do you think?

msz-rai (changes requested):
The flag you added will be handy. Please consider my suggestions.

Also, in your current implementation, `all` doesn't include ros2 extension with a standalone build.

setup.py
prybicki: Hmm, how about:?
```python
if args.with_all:
    args.with_pcl = True
    args.with_ros2 = True
```

PiotrMrozik: It much more clear then banch of ORs.

CMakeLists.txt
PiotrMrozik: I am aware of ros2standalone build.
This should be discussed. IMHO in "build all" mode we should choose "default" ros2 building method

setup.py
prybicki: @msz-rai Same applies to autotape, logging level, etc. These are indirectly available through `--cmake "-D...`. Not sure if it makes sense to lay pipes through setup to every option in cmake. I'd settle down only for the most important ones from user's point of view.

CMakeLists.txt
msz-rai: @prybicki, what are your thoughts?

CMakeLists.txt
```diff
@@ -99,7 +99,7 @@ add_library(RobotecGPULidar SHARED
 
 set_property(TARGET RobotecGPULidar PROPERTY POSITION_INDEPENDENT_CODE ON)
 
-if (RGL_BUILD_PCL_EXTENSION)
+if (RGL_BUILD_PCL_EXTENSION OR RGL_BUILD_ALL)
```
msz-rai, line 102: Leftovers after review changes. Please scan the whole CMakeLists.txt file.

setup.py
```diff
@@ -115,6 +122,7 @@ def is_cuda_version_ok():
     if on_windows():
         os.environ["Path"] = os.environ["Path"] + ";" + os.path.join(os.getcwd(), args.build_dir)
 
+
```
msz-rai, line 125:
```suggestion
```

setup.py
```diff
@@ -44,6 +44,8 @@ def main():
                         help="Install dependencies for PCL extension and exit")
     parser.add_argument("--clean-build", action='store_true',
                         help="Remove build directory before cmake")
+    parser.add_argument("--with-all", action='store_true',
+                        help="Build RGL with all possible extensions, tools and tests.")
```
msz-rai, line 48: At the moment, only extensions are handled with this flag. Add flags to enable building tools and tests. Or modify the help message.

msz-rai (changes requested):

prybicki:
At first sight, the suggestion seemed nice, however, it introduces ambiguities about what `all` means, e.g.,
- does it include ros2 standalone build?
- does it include tests and tools?
This flag can be replaced by typing individually desired `all` configuration and using reverse search in bash.

PR #122, opened 2023-03-27
Support multi-raytrace graphs

msz-rai:


src/graph/NodesCore.hpp
```diff
@@ -88,6 +88,9 @@ struct CompactPointsNode : Node, IPointsNodeSingleInput
 	void validate() override;
 	void schedule(cudaStream_t stream) override;
 
+	// Node requirements
+	std::vector<rgl_field_t> getRequiredFieldList() const override { return {IS_HIT_I32}; }
```
prybicki, line 92: Good point :)

src/graph/RaytraceNode.cpp
```diff
@@ -93,3 +95,31 @@ void RaytraceNode::setFields(const std::set<rgl_field_t>& fields)
...
+						outFields.insert(field);
+					}
+				}
+				dfsRet(node, dfsInputs);
+			}
+		}
+	};
+
+	dfsRet(shared_from_this(), true);  // Search in inputs
+	dfsRet(shared_from_this(), false);  // Search in outputs
```
prybicki, line 122: Please comment that the only case where input matters in SetRingIds (this node may change in the future).

src/graph/RaytraceNode.cpp
```diff
@@ -22,6 +22,8 @@
 
 void RaytraceNode::validate()
 {
+	setFields(findFieldsToCompute());
+
```
prybicki, line 26: It would be good to add a comment that it should be viewed as a temporary solution.

prybicki: Also - I'd need help to solve it in v14. `validate()` is replaced with a few functions:
- `onInputChange()`
- `onGraphChange()`
None is always called before execution as `validate()` was; they are meant to be called only when X changes. I could imagine a similar method for handling changing field requirements. However, it might get messy and tricky. Instead, I'd propose to move this burden onto user.

prybicki (commented):
Looks good, a few suggestions.

src/graph/RaytraceNode.cpp
msz-rai: Sure. Let's discuss this offline.

prybicki (approved):

PR #123, opened 2023-03-29
Make v13 build and pass tests on Windows

msz-rai:


prybicki (approved):
Bad whitespaces and unnecessarily removed prefix substring "Object does not exist", but overall looks good.

PR #136, opened 2023-04-06
Add UDP client

msz-rai:


src/UdpClient.hpp
```diff
@@ -0,0 +1,65 @@
...
+#ifdef _WIN32
+	#define SOCKET_TYPE SOCKET
+#else
+	#define SOCKET_TYPE int
+	#define INVALID_SOCKET (-1)
+#endif // _WIN32
+
+struct UdpClient
+{
+	// Ctor cannot throw, because we want Dtor call if instance creation failed
```
prybicki, line 43: Not only can they throw, but it is also a good practice. [Learn more here](https://isocpp.org/wiki/faq/exceptions#ctors-can-throw)
Destructor won't be called, but you can solve this by wrapping the constructor into [function-try-block](https://en.cppreference.com/w/cpp/language/function-try-block). When an exception is caught, call a private `cleanup()` method (also called by the destructor) to release all acquired resources.

src/UdpClient.cpp
```diff
@@ -0,0 +1,130 @@
...
+		return;
+	}
+	#endif
+
+	// Create socket
+	sockfd = socket(AF_INET, SOCK_DGRAM, 0);
+	if (sockfd == INVALID_SOCKET) {
+		initErrorMsg = fmt::format("cannot create socket: {}", LAST_ERROR);
+		return;
+	}
```
prybicki, line 62: Does it work on Windows? Does Winsock enable it? Please add some explanation why POSIX calls are sufficient here.

src/UdpClient.hpp
```diff
@@ -0,0 +1,65 @@
...
+	static UdpClient create(const std::string& srcAddr, int srcPort,
+	                        const std::string& dstAddr, int dstPort);
```
prybicki, line 45: Do we need to provide `srcPort`? According to [wiki](https://en.wikipedia.org/wiki/User_Datagram_Protocol) it is optional if the sender does not intend to handle incoming messages.

src/UdpClient.cpp
```diff
@@ -0,0 +1,130 @@
...
+	#define LAST_ERROR std::system_category().message(WSAGetLastError())
+	#define SO_BROADCAST_TYPE char
+#else
+	#define LAST_ERROR strerror(errno)
+	#define SO_BROADCAST_TYPE int
+#endif // _WIN32
+
+UdpClient UdpClient::createBroadcaster(const std::string& srcAddr, int srcPort)
+{
+	return create(srcAddr, srcPort, BROADCAST_ADDRESS, srcPort);
```
prybicki, line 29: Having the same src and dst port looks very suspicious to me.

src/UdpClient.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+	static UdpClient createBroadcaster(const std::string& srcAddr, int srcPort);
```
prybicki, line 47: Hmm, I'm surprised by `srcPort` here. It would make much more sense to provide only `dstPort`. Please fix/explain.

src/UdpClient.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+	~UdpClient();
+
+	void send(const std::vector<char>& buffer);
+
+private:
+	UdpClient(const std::string& srcAddr, int srcPort,
+	          const std::string& dstAddr, int dstPort);
+
+	bool addressTextToBinary(const std::string& textAddr, int port, sockaddr_in& outAddr);
```
prybicki, line 57: FYI: AFAIK using [string_view](https://en.cppreference.com/w/cpp/string/basic_string_view) is more idiomatic in C++17. But I'm totally OK with having `const std::string&` in this class. Let's not overengineer this.

prybicki, line 57: Just return sockaddr_in or throw.

src/UdpClient.hpp
```diff
@@ -0,0 +1,65 @@
...
+#include "RGLExceptions.hpp"
+
+#ifdef _WIN32
+	#define SOCKET_TYPE SOCKET
+#else
+	#define SOCKET_TYPE int
+	#define INVALID_SOCKET (-1)
+#endif // _WIN32
+
+struct UdpClient
```
prybicki, line 41: Is it really a client? A `sender`, `transmitter`, or `sink` sound better than `client`.

src/UdpClient.hpp
```diff
@@ -0,0 +1,65 @@
...
+{
+	// Ctor cannot throw, because we want Dtor call if instance creation failed
+	static UdpClient create(const std::string& srcAddr, int srcPort,
+	                        const std::string& dstAddr, int dstPort);
+
+	static UdpClient createBroadcaster(const std::string& srcAddr, int srcPort);
+
+	~UdpClient();
+
+	void send(const std::vector<char>& buffer);
```
prybicki, line 51: You will want to change this into, e.g., a raw pointer at some point because a copy from GPU to a vector is suboptimal.

prybicki (changes requested):
Please add simple GTests (for broadcast and unicast) proving that this class can send messages (receive them manually).

PR #137, opened 2023-04-09
Api call set instance

PiotrMrozik:


test/include/scenes.hpp
```diff
@@ -47,3 +47,26 @@ static inline void setupBoxesAlongAxes(rgl_scene_t scene)
...
+        rgl_mat3x4f boxCTf = Mat3x4f::TRS({ 6,  4,  2 }, { 0, 20, 60 }, { 1, 1, 1 }).toRGL();
+
+        EXPECT_RGL_SUCCESS(rgl_entity_set_pose(boxA, &boxATf));
+        EXPECT_RGL_SUCCESS(rgl_entity_set_pose(boxB, &boxBTf));
+        EXPECT_RGL_SUCCESS(rgl_entity_set_pose(boxC, &boxCTf));
+
+        EXPECT_RGL_SUCCESS(rgl_entity_set_id(boxA, 1));
+        EXPECT_RGL_SUCCESS(rgl_entity_set_id(boxB, 2));
+        EXPECT_RGL_SUCCESS(rgl_entity_set_id(boxC, 3));
+}
```
prybicki, line 72: Hmm, consider reusing `setupBoxesAlongAxes` just above.

test/src/features/instanceID.cpp
```diff
@@ -0,0 +1,70 @@
...
+    } formatStruct;
+
+    rgl_configure_logging(RGL_LOG_LEVEL_ALL, nullptr, true);
+    EXPECT_RGL_SUCCESS(rgl_node_rays_from_mat3x4f(&useRaysNode, rays.data(), rays.size()));
+    EXPECT_RGL_SUCCESS(rgl_node_raytrace(&raytraceNode, nullptr, 1000));
+    EXPECT_RGL_SUCCESS(rgl_node_points_compact(&compactNode));
+    EXPECT_RGL_SUCCESS(rgl_node_points_format(&formatNode, formatFields.data(), formatFields.size()));
+#ifdef RGL_BUILD_ROS2_EXTENSION
+
+#endif
```
prybicki, line 35: Leftover.

test/src/features/instanceID.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+    std::vector<rgl_field_t> formatFields = {
+        XYZ_F32,
+        ENTITY_IDX_I32
+    };
+    struct FormatStruct
+    {
+        Field<XYZ_F32>::type xyz;
+        Field<ENTITY_IDX_I32>::type entityId;
+    } formatStruct;
```
prybicki, line 26: You don't need this. The same effect (and less typing) can be achieved using YieldPointsNode (and getting results as SoAs).

test/src/features/instanceID.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+    rgl_configure_logging(RGL_LOG_LEVEL_ALL, nullptr, true);
```
prybicki, line 28: Leftover?

test/src/features/instanceID.cpp
```diff
@@ -0,0 +1,70 @@
...
+//#endif
+
+
+    // This test is super naive, but it's a start.
+    //TODO Make this test object sensitive: check that the xyz values are within the expected range for each object, and that compare its IDs.
+    for (int i = 0; i < formatData.size(); ++i)
+    {
+        EXPECT_GE(formatData[i].entityId, 1);
+        EXPECT_LE(formatData[i].entityId, 3);
+    }
```
prybicki, line 68: Yes, improve this.

prybicki (changes requested):

prybicki:
This PR could have a better name ;)

test/src/scene/instanceID.cpp
```diff
@@ -0,0 +1,79 @@
...
+//    EXPECT_RGL_SUCCESS(rgl_node_points_ros2_publish(&ros2publishNode, "pointcloud", "rgl"));
+//    EXPECT_RGL_SUCCESS(rgl_graph_node_add_child(formatNode, ros2publishNode));
+//    while(true)
+//        EXPECT_RGL_SUCCESS(rgl_graph_run(raytraceNode));
+//#endif
+
+	for (int i = 0; i < rays.size(); ++i) {
+
+		if (outIsHit[i]) {
+			// Check that the points are within the expected range for each object.
```
PiotrMrozik, line 61:
```suggestion
	for (int i = 0; i < rays.size(); ++i)
	{
		if (outIsHit[i]) {
			// Check that the points are within the expected range for each object.
```

test/src/scene/instanceID.cpp
```diff
@@ -0,0 +1,79 @@
...
+			}
+		}
+		else {
+			EXPECT_EQ(outID[i], 0);
+		}
+
+
+	}
+
+}
```
PiotrMrozik, line 79:
```suggestion
			}
		}
		else {
			EXPECT_EQ(outID[i], 0);
		}
	}
}
```

src/scene/Entity.hpp
```diff
@@ -20,6 +20,9 @@
 #include <utility>
 #include <math/Mat3x4f.hpp>
 
+constexpr int INVALID_ENTITY_ID = 0;
+constexpr int DEFAULT_ENTITY_ID = std::numeric_limits<int>::max();
```
prybicki, line 24: SSOT violation!

test/include/scenes.hpp
```diff
@@ -47,3 +47,26 @@ static inline void setupBoxesAlongAxes(rgl_scene_t scene)
...
+        rgl_mat3x4f boxCTf = Mat3x4f::TRS({ 6,  5, 0 }, { 0, 0, 0, }, { 1, 1, 1 }).toRGL();
+
+        EXPECT_RGL_SUCCESS(rgl_entity_set_pose(boxA, &boxATf));
+        EXPECT_RGL_SUCCESS(rgl_entity_set_pose(boxB, &boxBTf));
+        EXPECT_RGL_SUCCESS(rgl_entity_set_pose(boxC, &boxCTf));
+
+        EXPECT_RGL_SUCCESS(rgl_entity_set_id(boxA, 1));
+        EXPECT_RGL_SUCCESS(rgl_entity_set_id(boxB, 2));
+        EXPECT_RGL_SUCCESS(rgl_entity_set_id(boxC, 3));
+}
```
prybicki, line 72: After giving it a thought - I don't like the existence of this function. It hardcodes three, positions and sets instance IDs. Practically no reusability. I'd prefer to have a function taking needed args (with optional entity id) and calling it three times.

test/src/scene/instanceID.cpp
```diff
@@ -0,0 +1,76 @@
...
+			}
+			if (outPoints[i][1] > 2.0f) {
+				EXPECT_EQ(outID[i], 3);
+			}
+		}
+		else {
+			EXPECT_EQ(outID[i], 0);
+		}
+	}
+}
```
prybicki, line 76: You can also check that not setting entity ID results in DEFAULT_ENTITY_ID.

src/scene/Entity.hpp
```diff
@@ -20,19 +20,25 @@
...
+
 
 struct Entity : APIObject<Entity>
 {
 	Entity(std::shared_ptr<Mesh> mesh, std::optional<std::string> name=std::nullopt);
 
 	// TODO(prybicki): low-prio optimization: do not rebuild whole IAS if only transform changed
 	void setTransform(Mat3x4f newTransform);
+	void setId(int newId);
+	const int getId(){ return id; }
```
prybicki, line 34:
```suggestion
	const int getId() { return id; }
```

prybicki (changes requested):

test/include/scenes.hpp
PiotrMrozik: I will make a compromise

prybicki (approved):

PR #138, opened 2023-04-11
Asynchronous graph execution

prybicki:


test/src/synchronization/graphThreadSynchronization.cpp
```diff
@@ -0,0 +1,77 @@
...
+    }
+};
+
+INSTANTIATE_TEST_SUITE_P(graphThreadSynchronization, graphThreadSynchronization, testing::Values(1, 10, 1000));
+
+TEST_F(graphThreadSynchronization, LinearSingleRun)
+{
+    LinearGraph graph {};
+    graph.RunGraph();
+}
```
PiotrMrozik, line 57:
```suggestion

```

PiotrMrozik: With parameterized values, this test now is redundant.

test/src/synchronization/graphThreadSynchronization.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+TEST_F(graphThreadSynchronization, AddNodeAfterRun)
+{
+    LinearGraph graph {};
+    graph.RunGraph();
+
+    graph.addNodeAfterRun();
+
+    graph.RunGraph();
+}
```
PiotrMrozik, line 77:
```suggestion
}
// TODO(nebraszka): Write more unit test: trying to remove, add nodes when running is queued, accesing data, created, or not, etc.
```

src/graph/Node.hpp
```diff
@@ -131,6 +133,18 @@ struct Node : APIObject<Node>, std::enable_shared_from_this<Node>
 	 */
 	virtual void validateImpl() = 0;
 
+	/**
+	 * @return True, if node can be executed or queried for results.
```
PiotrMrozik, line 137: This may suggest, that we can query node without execution.

src/graph/GraphRunCtx.cpp
```diff
@@ -72,12 +114,71 @@ std::vector<std::shared_ptr<Node>> GraphRunCtx::findExecutionOrder(std::set<std:
...
+	maybeThread->join();
+	maybeThread.reset();
+}
+
+void GraphRunCtx::synchronizeNodeCPU(Node::Ptr nodeToSynchronize)
+{
+	if (!maybeThread.has_value()) {
+		return; // Already synchronized or never run.
+	}
+	// Wait until node is executed
```
PiotrMrozik, line 161: I would add comment about busy wait.

src/graph/Node.hpp
prybicki: Yes, this is utterly wrong.

msz-rai (approved):
Top implementation :1st_place_medal: 
I have left some minor comments to address. Anyway, approved!

src/graph/FromArrayPointsNode.cpp
```diff
@@ -37,7 +37,7 @@ void FromArrayPointsNode::setParameters(const void* points, size_t pointCount, c
 	auto gpuFields = gpuFieldDescBuilder.buildWritable(getFieldToPointerMappings(fields));
 	const char* inputPtr = static_cast<const char*>(inputData->getReadPtr(MemLoc::Device));
 	gpuFormatAosToSoa(nullptr, pointCount, pointSize, fields.size(), inputPtr, gpuFields->getWritePtr());
-	// TODO(msz-rai): check sync is necessary
+	// TODO(msz-rai): check synchronize is necessary
 	CHECK_CUDA(cudaStreamSynchronize(nullptr));
```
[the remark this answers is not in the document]
prybicki: Good idea, however the null stream is entirely removed in the following PRs.

src/graph/GraphRunCtx.hpp
```diff
@@ -33,27 +34,64 @@ struct GraphRunCtx
...
 private:
 	CudaStream::Ptr stream;
-	std::set<std::shared_ptr<Node>> nodes;
-	std::vector<std::shared_ptr<Node>> executionOrder;
+	std::optional<std::thread> maybeThread;
+	std::set<Node::Ptr> nodes;
+	std::vector<Node::Ptr> executionOrder;
+
+	// Used to synchronize all existing instances (e.g. to safely access Scene).
+	static std::list<std::weak_ptr<GraphRunCtx>> instances;
```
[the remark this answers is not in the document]
prybicki: Both solutions can be justified. I decided to move it to `Node` because it is frequently used there, and having a shared_ptr instead of weak_ptr removes all `weak_ptr::lock()` overhead.

PR #139, opened 2023-04-14
Fix transform points node test

nebraszka:
When testing `invalid_arguments` and `valid_arguments`, the `rgl_node_rays_transform` call was used instead of `rgl_node_points_transform`. I have discussed it with @PiotrMrozik.


PiotrMrozik (approved):

prybicki (approved):
Great idea to test the right API call :joy:

PR #140, opened 2023-04-14
Use graph stream in nodes

prybicki:
This PR contains mostly mundane refactoring to remove the passing stream to Node::executeImpl and use the stream from the bound GraphRunCtx.
It adds explicit syncs in VArray (to be removed...) to pass tests.


nebraszka (approved):

prybicki:
Merging to reduce PR queue. Changes are not critical here, so feel free to skip your review @PiotrMrozik .

PR #154, opened 2023-06-13
Q2/instance segmentation

PiotrMrozik:


test/include/scenes.hpp
```diff
@@ -24,26 +25,31 @@ static rgl_entity_t makeEntity(rgl_mesh_t mesh= nullptr, rgl_scene_t scene=nullp
 	return entity;
 }
 
-static inline void setupBoxesAlongAxes(rgl_scene_t scene)
+static inline void spawnCubeOnScene(rgl_scene_t scene,
+                                    const Vec3f &position = {0, 0, 0},
+                                    const Vec3f &rotation = {0, 0, 0},
+                                    const Vec3f &scale = {1, 1, 1},
```
msz-rai, line 31: It could be simplified by replacing these 3 parameters with `const Mat3x4f transform`

test/src/scene/instanceID.cpp
```diff
@@ -0,0 +1,80 @@
+#include "lidars.hpp"
+#include "scenes.hpp"
+#include "scene/Scene.hpp"
```
msz-rai, line 3: We should only use API headers in the tests if possible. It seems that it was included because of `DEFAULT_ENTITY_ID` variable. When my previous comment would be resolved this include can be removed.

test/include/scenes.hpp
```diff
@@ -4,6 +4,7 @@
 #include <models.hpp>
 #include <gtest/gtest.h>
 #include <utils.hpp>
+#include <scene/Entity.hpp>
```
msz-rai, line 7: ditto

test/src/scene/instanceID.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include "utils.hpp"
+
+#ifdef RGL_BUILD_ROS2_EXTENSION
+#include <rgl/api/extensions/ros2.h>
+#endif
+
+class InstanceIDTest : public RGLTest {
+};
+
+TEST_F(InstanceIDTest, BaseTest)
```
msz-rai, line 13: Please refactor this test. More details I can provide you offline.

msz-rai (commented):
Good job. It works :+1: 
Please address my comments.

test/src/scene/instanceID.cpp
PiotrMrozik: in this case, the best I can do is to delete "scene/Scene.hpp"

test/include/scenes.hpp
PiotrMrozik: I know that it is shorter this way, however, from the user's point of view 3 args are more convenient.
For example for parameterizations, quick changes of one vector and so one. In Your case, U have to create transform each time, and it will be a pain in the ass.

msz-rai: `Mat3x4f` has a nice set of methods to create a transform. I see no pain here.
`TRS` method has the same parameres as yours, so it could be used in the parameterizations.

include/rgl/api/core.h
```diff
@@ -45,6 +45,14 @@
 #define RGL_VERSION_MINOR 13
 #define RGL_VERSION_PATCH 1
 
+// Invalid entity ID is assign to rays that does not hit any entity.
+// Cannot be assigned to mesh manually. It is reserved for internal raytracing use.
+#define RGL_ENTITY_INVALID_ID  0
+
+// Default entity ID is the largest positive 32-bit integer.
+// It is assigned by defolut if the user does not specify it.
```
msz-rai, line 53:
```suggestion
// It is assigned by default if the user does not specify it.
```

include/rgl/api/core.h
```diff
@@ -339,6 +347,14 @@ rgl_entity_destroy(rgl_entity_t entity);
 RGL_API rgl_status_t
 rgl_entity_set_pose(rgl_entity_t entity, const rgl_mat3x4f *transform);
 
+/**
+ * Set instance ID of the given entity.
+ * @param entity Entity to modify
+ * @param int ID to set. If not set, value of entity id will be automatically generated a DEFAULT_ENTITY_ID = 2147483647.
```
msz-rai, line 353:
```suggestion
 * @param int ID to set. If not set, value of the entity id will be automatically generated as a DEFAULT_ENTITY_ID.
```

test/include/scenes.hpp
PiotrMrozik: Fair point. I forgot about TRS!
Jokes on me

test/include/scenes.hpp
```diff
@@ -24,26 +24,32 @@ static rgl_entity_t makeEntity(rgl_mesh_t mesh= nullptr, rgl_scene_t scene=nullp
 	return entity;
 }
 
-static inline void setupBoxesAlongAxes(rgl_scene_t scene)
+static inline void spawnCubeOnScene(rgl_scene_t scene,
+                                    Mat3x4f transform,
```
msz-rai, line 28:
```suggestion
                                    const Mat3x4f &transform,
```

msz-rai (approved):
LGTM :+1:

PR #168, opened 2023-07-13
Improve range configuration

msz-rai:


src/gpu/optixPrograms.cu
```diff
@@ -139,10 +136,20 @@ extern "C" __global__ void __closesthit__()
 		optixGetPayload_2()
 	});
 
+	float distance = sqrt(pow((hitWorld)[0] - (origin)[0], 2) +
```
PiotrMrozik, line 139: We can think of and discuss the little neat trick.
We can use Inversesqrt here, which is one operation cheaper then sqrt.
This will cost us passing inverse distance to the shader, however I think we should think about details like that on Our way to 1B rays/second.

src/graph/NodesCore.hpp
```diff
@@ -194,10 +193,12 @@ struct FromMat3x4fRaysNode : Node, IRaysNode
 
 	// Rays description
 	size_t getRayCount() const override { return rays->getCount(); }
+	size_t getRangesCount() const override { return 0; }  // To be set in SetRangeRaysNode
 	std::optional<size_t> getRingIdsCount() const override { return std::nullopt; }
 
 	// Data getters
 	VArrayProxy<Mat3x4f>::ConstPtr getRays() const override { return rays; }
+	VArrayProxy<Vec2f>::ConstPtr getRanges() const override { return nullptr; }  // To be set in SetRangeRaysNode
```
PiotrMrozik, line 201: We assume that this will be set always?
If not, then optional.

src/graph/RaytraceNode.cpp
```diff
@@ -27,6 +27,11 @@ void RaytraceNode::validate()
 
 	raysNode = getValidInput<IRaysNode>();
 
+	if (raysNode->getRanges() == nullptr || raysNode->getRangesCount() == 0) {
```
PiotrMrozik, line 30: As I think about this, I rly do not like the idea, that we have a node (which logically is optional), but in reality is required in the pipeline. What if I wanna simulate some rays (abstract, not lidar), and I do not have/need ranges? 
this requirement seems artificial.

src/graph/SetRangeRaysNode.cpp
```diff
@@ -0,0 +1,47 @@
...
+void SetRangeRaysNode::setParameters(const Vec2f* rangesRaw, size_t rangesCount)
+{
+	// Validate ranges
+	for (int i = 0; i < rangesCount; ++i) {
+		float minRange = rangesRaw[i][0];
+		float maxRange = rangesRaw[i][1];
+		if (std::isnan(minRange) || std::isnan(maxRange)) {
+			throw InvalidAPIArgument("range must be a number");
+		}
+		if (minRange > maxRange) {
```
PiotrMrozik, line 26: `>=`

test/src/DistanceFieldTest.cpp
```diff
@@ -74,17 +74,71 @@ TEST_F(DistanceFieldTest, should_compute_infinite_distance_for_object_off_ray_pa
...
+    connectNodes(false);
 
     getResults();
 
+    ASSERT_EQ(outDistancesCount, rayTf.size());
     EXPECT_EQ(outDistances.at(0), NON_HIT_VALUE);
     EXPECT_EQ(outDistances.at(1), NON_HIT_VALUE);
 }
 
+TEST_P(DistanceFieldTest, should_compute_infinite_distance_for_object_out_of_max_range)
```
PiotrMrozik, line 87:
```suggestion
TEST_P(DistanceFieldTest, out_of_max_range_case)
```

test/src/DistanceFieldTest.cpp
```diff
@@ -74,17 +74,71 @@ TEST_F(DistanceFieldTest, should_compute_infinite_distance_for_object_off_ray_pa
...
+    prepareNodes({0.0f, cubeZDistance - CUBE_HALF_EDGE - 1.0f});
+    connectNodes(false);
+
+    getResults();
+
+    ASSERT_EQ(outDistancesCount, rayTf.size());
+    EXPECT_EQ(outDistances.at(0), NON_HIT_VALUE);
+}
+
+TEST_P(DistanceFieldTest, should_compute_infinite_distance_for_object_out_of_min_range)
```
PiotrMrozik, line 104:
```suggestion
TEST_P(DistanceFieldTest, out_of_min_range_case)
```

test/src/DistanceFieldTest.cpp
```diff
@@ -74,17 +74,71 @@ TEST_F(DistanceFieldTest, should_compute_infinite_distance_for_object_off_ray_pa
...
+    prepareNodes({cubeZDistance + CUBE_HALF_EDGE + 1.0f, cubeZDistance + CUBE_HALF_EDGE + 2.0f});
+    connectNodes(false);
+
+    getResults();
+
+    ASSERT_EQ(outDistancesCount, rayTf.size());
+    EXPECT_EQ(outDistances.at(0), NON_HIT_VALUE);
+}
+
+TEST_P(DistanceFieldTest, should_compute_infinite_distance_for_object_in_range_behind_object_out_of_min_range)
```
PiotrMrozik, line 121: ditto

PiotrMrozik (changes requested):

include/rgl/api/core.h
```diff
@@ -469,6 +469,21 @@ rgl_node_rays_from_mat3x4f(rgl_node_t* node, const rgl_mat3x4f* rays, int32_t ra
...
+ * Creates or modifies SetRangeRaysNode.
+ * The node defines min and max range for existing rays.
+ * It could be passed:
+ * - one range value and it will be applied to all of the rays
+ * - number of ranges equals to the number of rays (range for each ray will be specified individually)
+ * Input: rays
+ * Output: rays
+ * @param node If (*node) == nullptr, a new node will be created. Otherwise, (*node) will be modified.
+ * @param ranges Pointer to ranges.
+ * @param ranges_count Size of the `ranges` array.
```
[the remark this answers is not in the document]
msz-rai: I am following the style used in other API calls documentation.

src/gpu/optixPrograms.cu
msz-rai: Good point for the optimization task. I will leave a TODO comment here.

src/graph/SetRangeRaysNode.cpp
msz-rai: I see no reason that couldn't be equal

test/src/DistanceFieldTest.cpp
nebraszka: @PiotrMrozik Why? I think this name better describes what the test checks

src/graph/SetRangeRaysNode.cpp
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	ranges->setData(rangesRaw, rangesCount);
+}
+
+void SetRangeRaysNode::validate()
+{
+	input = getValidInput<IRaysNode>();
+
+	if (!(ranges->getCount() == 1 || ranges->getCount() == input->getRayCount())) {
```
PiotrMrozik, line 41: I'd like to know if you can merge those two cases. Just hold 1 as a number of rays ;)

PiotrMrozik (approved):

PR #173, opened 2023-08-08
Add target triplet of vcpkg to cmake

msz-rai:


Jakub-Krakowiak (approved):
It fixes errors on my machine.

PR #174, opened 2023-08-11
Remove MemoryKind tparam from IAnyArray and below

prybicki:


src/memory/Array.hpp
```diff
@@ -33,12 +33,12 @@
  * @tparam M Allows to statically distinguish Arrays using different memory kinds.
  * @tparam T Type of elements
  */
-template<MemoryKind M, typename T>
-struct Array : public IAnyArray<M>
+template<typename T>
+struct Array : public IAnyArray
```
PawelLiberadzki, line 37: M template parameter not removed from documentation.
There are also duplicated headers near the top of this file.

src/memory/ArrayProxy.hpp
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#include <memory/IAnyArray.hpp>
+
+struct ArrayHub
```
PawelLiberadzki, line 5: This class seems to be not used anywhere and should probably be moved out of this request.

PawelLiberadzki (approved):

PR #175, opened 2023-08-16
Fix linking tests on Windows by exporting symbols

msz-rai:
Some tests on Windows were not linked properly. It was because Windows does not export all symbols in a shared library automatically ([see](https://www.kitware.com//create-dlls-on-windows-without-declspec-using-new-cmake-export-all-feature/)).
This fix allowed to remove `Only Linux` section in test CMakeLists.

### DIFF
--- a/test/src/tapeSurfaceTest.cpp
+++ b/test/src/tapeSurfaceTest.cpp
@@ -22,18 +22,18 @@ TEST_F(TapeCase, RecordPlayLoggingCall)
 	std::string logFilePath { (std::filesystem::temp_directory_path() / std::filesystem::path("loggingRecord").concat(".log")).string() };
 
 	// Logging with Tape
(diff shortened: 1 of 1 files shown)


test/src/tapeSurfaceTest.cpp
```diff
@@ -16,9 +16,30 @@
 
 class TapeCase : public RGLTest {};
 
+TEST_F(TapeCase, RecordPlayLoggingCall)
+{
+	std::string loggingRecordPath { (std::filesystem::temp_directory_path() / std::filesystem::path("loggingRecord")).string() };
+	std::string logFilePath { (std::filesystem::temp_directory_path() / std::filesystem::path("loggingRecord").concat(".log")).string() };
+
+	// Logging with Tape
```
nebraszka, line 24: mismatched indentation

nebraszka (approved):

PR #178, opened 2023-08-29
Format points node tests

nebraszka:


test/src/graph/nodes/FormatPointsNodeTest.cpp
```diff
@@ -1,39 +1,232 @@
...
+	std::vector<rgl_node_t> formatNodes(MULTIPLE_FORMATS_COUNT);
+	ASSERT_RGL_SUCCESS(rgl_node_points_format(&formatNodes.at(0), fieldsToFormat.data(), fieldsToFormat.size()));
+	ASSERT_THAT(formatNodes.at(0), testing::NotNull());
+
+	ASSERT_RGL_SUCCESS(rgl_graph_node_add_child(inNode, formatNodes.at(0)));
+	for (int i = 1; i < MULTIPLE_FORMATS_COUNT; ++i) {
+		ASSERT_RGL_SUCCESS(rgl_node_points_format(&formatNodes.at(i), fieldsToFormat.data(), fieldsToFormat.size()));
+		ASSERT_THAT(formatNodes.at(i), testing::NotNull());
+		ASSERT_RGL_SUCCESS(rgl_graph_node_add_child(formatNodes.at(i - 1), formatNodes.at(i)));
+	}
```
prybicki, line 137: IMHO format nodes between the first and the last could have different fields.

prybicki: Oh, there's such a test. Do you think there's any benefit of keeping this version?

test/src/graph/nodes/FormatPointsNodeTest.cpp
```diff
@@ -1,39 +1,232 @@
...
+		TestPointCloud pointCloud = TestPointCloud(pointFields, pointsCount);
+		inNode = pointCloud.createUsePointsNode();
+
+		ASSERT_RGL_SUCCESS(rgl_node_points_format(&formatNode, allNotDummyFields.data(), allNotDummyFields.size()));
+		ASSERT_RGL_SUCCESS(rgl_graph_node_add_child(inNode, formatNode));
+		EXPECT_RGL_INVALID_PIPELINE(rgl_graph_run(inNode), fmt::format("requires {} to be present", toString(fieldToRemove)));
+
+		pointFields.emplace_back(fieldToRemove);
+		rgl_graph_node_remove_child(inNode, formatNode);
+	}
```
prybicki, line 231: :1st_place_medal:

prybicki (changes requested):
Great tests. A number of suggestions + one critical bug.

test/include/helpers/commonHelpers.hpp
```diff
@@ -25,8 +25,8 @@ static constexpr int maxGPUCoresTestCount = 20000;
 	} while (false)
 
 #define EXPECT_RGL_INVALID_OBJECT(status, type) EXPECT_RGL_STATUS(status, RGL_INVALID_API_OBJECT, "Object does not exist", type)
-#define EXPECT_RGL_INVALID_ARGUMENT(status, error) EXPECT_RGL_STATUS(status, RGL_INVALID_ARGUMENT, "Invalid argument", error)
-#define EXPECT_RGL_INVALID_PIPELINE(status, error) EXPECT_RGL_STATUS(status, RGL_INVALID_PIPELINE, "")
+#define EXPECT_RGL_INVALID_ARGUMENT(status, error) EXPECT_RGL_STATUS(status, RGL_INVALID_ARGUMENT, error)
+#define EXPECT_RGL_INVALID_PIPELINE(status, error) EXPECT_RGL_STATUS(status, RGL_INVALID_PIPELINE, error)
```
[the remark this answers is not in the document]
nebraszka: Because it does not always appear as a message in this err:
`throw InvalidAPIArgument("cannot format field 'RGL_FIELD_DYNAMIC_FORMAT'");`

test/src/graph/nodes/FormatPointsNodeTest.cpp
nebraszka: Perhaps it would make more sense to check that **the same** format used repeatedly does not affect the data. But as far as I know RGL does not allow you to create such a graph.

test/include/helpers/testPointCloud.hpp
```diff
@@ -151,90 +249,97 @@ class TestPointCloud
...
+				std::move(reinterpret_cast<char*>(fieldValues.data() + i), reinterpret_cast<char*>(fieldValues.data() + i + 1),
+				          data.data() + i * getPointByteSize() + offset);
+			}
 		}
 	}
 
 	template<rgl_field_t T>
 	std::vector<typename Field<T>::type> getFieldValues()
 	{
+		int fieldIndex = std::find(fields.begin(), fields.end(), T) - fields.begin();
```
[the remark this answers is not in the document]
nebraszka: >Possibly ambiguous behavior if isDummy(T).

Why?

I assume that if there are multiple of the same field, they have the same value - I don't see (nomen omen) the point of a point that has, for example, two different isHit values.

test/include/helpers/commonHelpers.hpp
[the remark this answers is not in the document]
prybicki: I'd fix this throw, instead of deleting check for all other tests.

prybicki (approved):

PR #181, opened 2023-09-13
Fix missing dependencies for ROS2 standalone build

msz-rai:
Based on: https://github.com/RobotecAI/ros2cs/pull/54


prybicki (approved):
Talked through it in person.

PR #182, opened 2023-09-19
Fix race condition related to getFieldData in lazy nodes

prybicki:


src/CacheManager.hpp
```diff
@@ -48,6 +48,15 @@ struct CacheManager
 		cacheAge.erase(key);
 	}
 
+	std::vector<Key> getKeys() {
```
PawelLiberadzki, line 51: This method may be const.

src/api/apiCore.cpp
```diff
@@ -535,8 +535,15 @@ void TapePlayer::tape_graph_get_result_size(const YAML::Node& yamlNode)
 		&out_count,
 		&out_size_of);
 
-	if (out_count != yamlNode[2].as<int32_t>()) RGL_WARN("tape_graph_get_result_size: out_count mismatch");
-	if (out_size_of != yamlNode[3].as<int32_t>()) RGL_WARN("tape_graph_get_result_size: out_size_of mismatch");
+	auto tape_count = yamlNode[2].as<int32_t>();
```
PawelLiberadzki, line 538: Naming convention seem to not match a lot of code in this file: snake -> camel?
+
Also, tape_count and tape_size_of may be const.

prybicki:
Fixed/answered, waiting for approve.

src/api/apiCore.cpp
PawelLiberadzki: Regarding the naming convention, understandable. In terms of const correctness, it is rather not a matter of performace, that is right. It is more about semantics - if something should not be mutable, we simply make it const.

PawelLiberadzki (approved):

PR #183, opened 2023-09-19
Improve tape tools to handle multi-lidar tapes

prybicki:


tools/tapeVisualizer.cpp
```diff
@@ -18,33 +18,130 @@
 #include "Tape.hpp"
 #include "macros/checkRGL.hpp"
 
-int main(int argc, char** argv)
+constexpr std::string_view INTERCEPTED_NODE_CALL = "rgl_node_points_compact";
+constexpr size_t FIELDS_COUNT = 1;
+constexpr rgl_field_t FIELDS[FIELDS_COUNT] = {RGL_FIELD_XYZ_F32 };
```
msz-rai, line 23: I'd move these variables close to `std::vector<rgl_vec3f> points`, because their modification has an impact on the container. Also, for now, only one field is supported.
Or put a comment to not modify these variables.

tools/tapeVisualizer.cpp
msz-rai, line 1: `tapeVisualizer` works well with multiple LiDARs that have the same capture rate. I think this note is worth mentioning.

msz-rai (approved):
LGTM

tools/tapeVisualizer.cpp
prybicki: It has an impact on spatial merge too. Therefore, not moving closer, but adding a comment not to modify.

PR #191, opened 2023-09-25
Implement rgl_graph_node_set_priority

prybicki:


include/rgl/api/core.h
```diff
@@ -698,3 +698,13 @@ rgl_graph_node_add_child(rgl_node_t parent, rgl_node_t child);
  */
 RGL_API rgl_status_t
 rgl_graph_node_remove_child(rgl_node_t parent, rgl_node_t child);
+
+/**
+ * Allows to set relative node's priority, which determine their execution order.
+ * This is useful when some branch (e.g. visualization data)
+ * needs to be executed before some other (e.g. publishing to ROS2).
+ * By default, all nodes have priority 0.
+ * It is illegal to set node's priority to a value lower than max priority value of its outputs.
```
PawelLiberadzki, line 707: Other functions here have full documentation, with @param tags. Just for consistency, I would consider updating this dox.

src/graph/Node.cpp
```diff
@@ -180,3 +187,26 @@ cudaStream_t Node::getStreamHandle()
...
+	// It is illegal to set node's priority to a value lower than max priority value of its outputs.
+	if (!outputs.empty() && outputs.front()->priority > requestedPriority) {
+		auto msg = fmt::format("cannot set priority of node {} to {}, because it's children {} has higher priority {}",
+		                       getName(), requestedPriority, outputs.front()->getName(), outputs.front()->priority);
+		throw InvalidPipeline(msg);
+	}
+
+	this->priority = requestedPriority;
+	// Update parent nodes.
+	for (auto&& input : inputs) {
```
PawelLiberadzki, line 202: I thought about this and it may be different from your intentions, but I will write this down, at least as a concept.

Although the above should work, maybe there is an option to implement simplier concept - where node prio is resolved only between outputs? In other words, node is always resolved before its children, and then we decide based on priority.

prybicki: Discussed offline and decided to go with the current implementation (global priorities).

msz-rai (commented):
`rgl_graph_node_set_priority` can change the priority of the other nodes implicitly. I'd consider adding an API call to get the node's priority. It could help in bumping priority in runtime or for debugging purposes (on the plugin site).

src/graph/Node.cpp
```diff
@@ -49,6 +49,13 @@ void Node::addChild(Node::Ptr child)
 	this->outputs.push_back(child);
 	child->inputs.push_back(shared_from_this());
 
+	// Sort outputs by priority
+	std::stable_sort(outputs.begin(), outputs.end(),
+	                 [](Node::Ptr lhs, Node::Ptr rhs)
+	                 { return lhs->priority > rhs->priority; });
```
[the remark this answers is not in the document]
prybicki: Sometimes I sort ascending, sometimes descending, sometimes just a subset. IMHO does not justify adding another method.

src/api/apiCore.cpp
```diff
@@ -633,6 +633,27 @@ void TapePlayer::tape_graph_node_remove_child(const YAML::Node& yamlNode)
...
+		Node::validatePtr(node)->setPriority(priority);
+	});
+	TAPE_HOOK(node, priority);
+	return status;
+}
+
+void TapePlayer::tape_graph_node_set_priority(const YAML::Node& yamlNode)
+{
+	auto nodeId = yamlNode[0].as<TapeAPIObjectID>();
+	rgl_node_t node = tapeNodes.contains(nodeId) ? tapeNodes.at(nodeId) : nullptr;
```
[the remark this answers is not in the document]
prybicki: Discussed offline, rejected.

src/graph/Node.hpp
```diff
@@ -107,6 +107,9 @@ struct Node : APIObject<Node>, std::enable_shared_from_this<Node>
 	 */
 	std::set<Node::Ptr> disconnectConnectedNodes();
 
+	void setPriority(int32_t);
+	int32_t getPriority() { return priority; }
```
PawelLiberadzki, line 111: I would make this const method.

PawelLiberadzki (approved):

PR #203, opened 2023-11-02
Fixes around Entity (see description)

prybicki:
- Refactor and cleanup struct Entity
- Add documentation to Entity
- Move Entity::getIAS code to Scene::buildSBT
- Move Entity ID from SBT to OptixInstance (more efficient)
- Minor improvement to EntityIdTest

### DIFF
--- a/src/scene/Entity.cpp
+++ b/src/scene/Entity.cpp
@@ -33,7 +33,7 @@ void Entity::setTransform(Mat3x4f newTransform)
 
 void Entity::setId(int newId)
 {
-	auto optixMaxInstanceId = 1 << 28; // https://raytracing-docs.nvidia.com/optix7/guide/index.html#limits#limits
+	constexpr auto optixMaxInstanceId = 1 << 28; // https://raytracing-docs.nvidia.com/optix7/guide/index.html#limits#limits
 	if (newId >= optixMaxInstanceId) {
 		auto msg = fmt::format("Entity ID ({}) must be less than {}", newId, optixMaxInstanceId);
 		throw std::invalid_argument(msg);
--- a/src/scene/Entity.cpp
+++ b/src/scene/Entity.cpp
@@ -39,7 +39,7 @@ void Entity::setId(int newId)
 		throw std::invalid_argument(msg);
 	}
 	id = newId;
-	scene->requestSBTRebuild();
+	scene->requestASRebuild();
 }
 
 void Entity::setIntensityTexture(std::shared_ptr<Texture> texture)
--- a/src/scene/Entity.cpp
+++ b/src/scene/Entity.cpp
@@ -16,9 +16,14 @@
 
 API_OBJECT_INSTANCE(Entity);
 
-Entity::Entity(std::shared_ptr<Mesh> mesh, std::optional<std::string> name)
-  : mesh(std::move(mesh)), transform(Mat3x4f::identity()), humanReadableName(std::move(name)), id(RGL_DEFAULT_ENTITY_ID)
-{}
+std::shared_ptr<Entity> Entity::create(std::shared_ptr<Mesh> mesh, std::shared_ptr<Scene> scene)
+{
+	auto entity = APIObject<Entity>::create(mesh);
+	scene->addEntity(entity);
+	return entity;
+}
+
+Entity::Entity(std::shared_ptr<Mesh> mesh) : mesh(std::move(mesh)) {}
 
 void Entity::setTransform(Mat3x4f newTransform)
 {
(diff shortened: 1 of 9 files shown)


src/scene/Entity.cpp
```diff
@@ -28,26 +33,19 @@ void Entity::setTransform(Mat3x4f newTransform)
 
 void Entity::setId(int newId)
 {
+	auto optixMaxInstanceId = 1 << 28; // https://raytracing-docs.nvidia.com/optix7/guide/index.html#limits#limits
```
PawelLiberadzki, line 36: This may be const.

src/scene/Entity.hpp
```diff
@@ -14,37 +14,68 @@
...
+	 * @return Estimated velocity based on current and previous transforms.
+	 */
+	Mat3x4f getVelocity() const;
 
-	std::shared_ptr<Mesh> mesh;
-	std::shared_ptr<Texture> intensityTexture = nullptr;
+	/**
+	 * @return The Scene in which this Entity is present.
+	 */
+	std::shared_ptr<Scene> getScene() const { return scene; }
```
PawelLiberadzki, line 62: Technically scene returned this way is mutable, in other words, you have access to internal state of Entity. This could not matter that much (or at all), so please decide yourself, where keep const here or not.

test/src/scene/entityIdTest.cpp
```diff
@@ -94,13 +94,19 @@ TEST_F(EntityIdTest, BaseTest)
...
-	//    rgl_node_t ros2publishNode = nullptr;
-	//
-	//    EXPECT_RGL_SUCCESS(rgl_node_points_ros2_publish(&ros2publishNode, "pointcloud", "rgl"));
-	//    EXPECT_RGL_SUCCESS(rgl_graph_node_add_child(formatNode, ros2publishNode));
-	//    while(true)
-	//        EXPECT_RGL_SUCCESS(rgl_graph_run(raytraceNode));
-	//#endif
+// Infinite loop to publish point cloud to ROS2 topic for visualization purposes. Uncomment to use. Use wisely.
+#ifdef RGL_BUILD_ROS2_EXTENSION
+//	rgl_node_t formatNode = nullptr;
```
PawelLiberadzki, line 99: Why did you put commented out macro here? Is this maybe by mistake?

src/scene/Entity.cpp
prybicki: Even `constexpr` IMHO

src/scene/Entity.hpp
prybicki: Modifying `Scene` does not impact `Entity` in any way. `Scene` is kept in `Entity` only as a reference, not as its part. This reference cannot be changed or removed, therefore `std::shared_ptr<Scene> getScene() const` looks like a good decision to me. WDYT?

test/src/scene/entityIdTest.cpp
prybicki: No significant reason, this style looked better to me, as the #macro creates a kind of a "box" for the commented-out code, which looks more pleasing to me :shrug:. And it's less to uncomment later, #minimalism

src/scene/Entity.hpp
PawelLiberadzki: Fair point, you are right, this passing ptr to common scene, so no internal state is exposed. Definitely may stay const.

src/scene/Entity.hpp
```diff
@@ -14,37 +14,68 @@
...
+#include <RGLFields.hpp>
 #include <scene/Scene.hpp>
 #include <scene/Mesh.hpp>
-#include <APIObject.hpp>
-#include <utility>
 #include <math/Mat3x4f.hpp>
 
-
+/**
+ * Entity represents an instance of a mesh on the scene, with attributes such as transform, id, intensity texture, etc.
```
PiotrMrozik, line 26: This description is misleading. 
More of a "tuple". 
Phrasing this way suggests that mesh has the following attributes which is not true.

test/include/helpers/sceneHelpers.hpp
```diff
@@ -26,7 +26,7 @@ static rgl_entity_t makeEntity(rgl_mesh_t mesh = nullptr)
 	return entity;
 }
 
-static inline void spawnCubeOnScene(const Mat3x4f& transform, std::optional<int> id = std::nullopt)
+static inline rgl_entity_t spawnCubeOnScene(const Mat3x4f& transform, std::optional<int> id = std::nullopt)
```
PiotrMrozik, line 29: Any particular reason for that? 
If U return entity with cube, then name `spawnCubeOnScene` is little bit unprecise.

test/src/scene/entityIdTest.cpp
```diff
@@ -38,9 +38,9 @@ TEST_F(EntityIdTest, BaseTest)
 	constexpr float BOX1_BOX2_BORDER = (BOX1_Y_POS + BOX2_Y_POS) / 2.0f;
 	constexpr float BOX2_BOX3_BORDER = (BOX2_Y_POS + BOX3_Y_POS) / 2.0f;
 
-	spawnCubeOnScene(Mat3x4f::TRS({6, BOX1_Y_POS, 0}, {0, 0, 0}, {1, 1, 1}), BOX1_ID);
-	spawnCubeOnScene(Mat3x4f::TRS({6, BOX2_Y_POS, 0}, {0, 0, 0}, {1, 1, 1}), BOX2_ID);
-	spawnCubeOnScene(Mat3x4f::TRS({6, BOX3_Y_POS, 0}, {0, 0, 0}, {1, 1, 1}));
+	auto box1 = spawnCubeOnScene(Mat3x4f::TRS({6, BOX1_Y_POS, 0}, {0, 0, 0}, {1, 1, 1}), BOX1_ID);
```
PiotrMrozik, line 41: ditto

PiotrMrozik (approved):
LGTM!
some smol comments to consider

PawelLiberadzki (approved):

src/scene/Entity.hpp
prybicki: Good catch, fixed

test/src/scene/entityIdTest.cpp
prybicki: It is used in the commented-out text.

test/include/helpers/sceneHelpers.hpp
PiotrMrozik: now this is spawnAndGet, but maybe I'm to sensitive

PR #206, opened 2023-11-07
Compute absolute point velocity

prybicki:


src/gpu/helpersKernels.cu
```diff
@@ -30,4 +30,18 @@ void gpuSetupRandomNumberGenerator(cudaStream_t stream, size_t elementsCount, un
...
 	run(kSetupRandomNumberGenerator, stream, elementsCount, seed, outPHILOXStates);
-}
\ No newline at end of file
+}
+
+__global__ void kUpdateVertices(size_t vertexCount, Vec3f* newVerticesToVelocity, Vec3f* oldToNewVertices)
+{
+	LIMIT(vertexCount);
+	// See Mesh::updateVertices to understand the logic here.
+	Vec3f newVertex = newVerticesToVelocity[tid];
```
PiotrMrozik, line 39: Is` newVertex ` var is necessary?

src/gpu/optixPrograms.cu
```diff
@@ -163,21 +166,49 @@ extern "C" __global__ void __closesthit__()
...
+	if (ctx.sceneDeltaTime > 0) {
+		if (entityData.hasPrevFrameLocalToWorld) {
+			// Computing hit point velocity in simple words:
+			// From raytracing, we get hit point in Entity's coordinate frame (hitObject).
+			// Think of it as a marker dot on the Entity.
+			// Having access to Entity's previous pose, we can compute (entityData.prevFrameLocalToWorld * hitObject),
+			// where the marker dot would be in the previous raytracing frame (displacementOriginWorld).
+			// Then, we can connect marker dot in previous raytracing frame with its current position and obtain transformDisplacement vector
+			// Dividing transformDisplacement by time elapsed from the previous raytracing frame yields velocity vector.
+			Vec3f displacementOriginWorld = entityData.prevFrameLocalToWorld * hitObject;
```
PiotrMrozik, line 195: `displacementOriginWorld ` is very misleading.
This is just globalHitpointPosition. Not displacement

PiotrMrozik: or worldHitpointPosition if U want.

src/gpu/optixPrograms.cu
```diff
@@ -163,21 +166,49 @@ extern "C" __global__ void __closesthit__()
...
+
+	Vec3f velocity{NAN};
+	Vec3f transformDisplacement = {0, 0, 0};
+	Vec3f skinningDisplacement = {0, 0, 0};
+	if (ctx.sceneDeltaTime > 0) {
+		if (entityData.hasPrevFrameLocalToWorld) {
+			// Computing hit point velocity in simple words:
+			// From raytracing, we get hit point in Entity's coordinate frame (hitObject).
+			// Think of it as a marker dot on the Entity.
+			// Having access to Entity's previous pose, we can compute (entityData.prevFrameLocalToWorld * hitObject),
```
PiotrMrozik, line 191: Actually it is not `FrameLocalToWorld`
This is just model (or local if U want) transform.
In that case they are the same, but only cuz parent of the entity is world origin. There is no real parenting.
When we will have parent and child in global space, then local transform of child is not equal to global transform of it, which is in that case.
Let's use proper name of frame to avoid misleading.

All the names like `localToWorld` or `someToOther`, suggest that they are not "real transform", but matrix of base change.
Those are two different concepts.
In simple scenario when You have only one transform ("position" of the object in the scene), it's somehow working, but in absolutly every other this will be a big mess, that we will not dig up from.

src/gpu/optixPrograms.cu
```diff
@@ -163,21 +166,49 @@ extern "C" __global__ void __closesthit__()
...
+			// where the marker dot would be in the previous raytracing frame (displacementOriginWorld).
+			// Then, we can connect marker dot in previous raytracing frame with its current position and obtain transformDisplacement vector
+			// Dividing transformDisplacement by time elapsed from the previous raytracing frame yields velocity vector.
+			Vec3f displacementOriginWorld = entityData.prevFrameLocalToWorld * hitObject;
+			transformDisplacement = hitWorld - displacementOriginWorld;
```
PiotrMrozik, line 196: Also this.
`transformDisplacement` is shaggy.
This is just a change of position between frames: displacement. OR transform containing displacement.
transformDisplacement suggests one of two things:
a) this is a change of whole transform which is bigger dimension problem, then concept that we are dealing with
or
b) just a simple displacement, but in a form of 3x4 matrix containg only linear part (for example for convinience).
Both are not true.

src/gpu/optixPrograms.cu
```diff
@@ -163,21 +166,49 @@ extern "C" __global__ void __closesthit__()
...
+		// Some entities may have skinned meshes - in this case entity.vertexDisplacementSincePrevFrame will be non-null
+		bool wasSkinned = entityData.vertexDisplacementSincePrevFrame != nullptr;
+		if (wasSkinned) {
+			const Vec3f& vA = entityData.vertexDisplacementSincePrevFrame[triangleIndices.x()];
+			const Vec3f& vB = entityData.vertexDisplacementSincePrevFrame[triangleIndices.y()];
+			const Vec3f& vC = entityData.vertexDisplacementSincePrevFrame[triangleIndices.z()];
+			skinningDisplacement = Vec3f((1 - u - v) * vA + u * vB + v * vC);
+		}
+
+		velocity = (transformDisplacement + skinningDisplacement) / static_cast<float>(ctx.sceneDeltaTime);
```
PiotrMrozik, line 208: This whole abomination + skinning part would be just one matrix multiplication, if we will do skinning in RGL properly.
I would leave TODO here, and come back to this topic again... 
In scenes with a lot of NPC, we rly lose a lot...

src/scene/Scene.hpp
```diff
@@ -56,12 +56,18 @@ struct Scene
...
+	}
+	std::optional<Time> getTime() const { return time; }
+	std::optional<Time> getPrevTime() const { return prevTime; }
+	std::optional<Time> getDeltaTime() const { return prevTime.has_value() ? std::optional(*time - *prevTime) : std::nullopt; }
 
-	std::size_t getObjectCount();
+	std::size_t getObjectCount() const;
 
-	CudaStream::Ptr getStream();
+	CudaStream::Ptr getStream() const;
```
PawelLiberadzki, line 70: This method gives access to stream, that is a private member of Scene - consider not exposing it. Unless it is driven by other design choices obviously.

src/gpu/optixPrograms.cu
PiotrMrozik: `DisplacementVectorOrigin`

src/scene/Entity.cpp
```diff
@@ -48,4 +50,24 @@ void Entity::setIntensityTexture(std::shared_ptr<Texture> texture)
 	Scene::instance().requestSBTRebuild();
 }
 
-Mat3x4f Entity::getVelocity() const { throw std::runtime_error("unimplemented"); }
+std::optional<Mat3x4f> Entity::getPreviousFrameTransform() const
```
PiotrMrozik, line 53: `getPreviousFrameTransform` ->`getPreviousFrameLocalToWorldTransform`

src/scene/Scene.hpp
PawelLiberadzki: I do not have better solution with std::shared_ptr - we would like to restrict access but also use a tool for sharing. In such case, you can e.g. return const ref to the object itself. However, I assume that this would not be convenient here, because of previous design choices.

PawelLiberadzki (approved):

src/gpu/helpersKernels.cu
```diff
@@ -30,4 +30,18 @@ void gpuSetupRandomNumberGenerator(cudaStream_t stream, size_t elementsCount, un
                                    curandStatePhilox4_32_10_t* outPHILOXStates)
 {
 	run(kSetupRandomNumberGenerator, stream, elementsCount, seed, outPHILOXStates);
-}
\ No newline at end of file
+}
+
+__global__ void kUpdateVertices(size_t vertexCount, Vec3f* newVerticesToVelocity, Vec3f* oldToNewVertices)
```
msz-rai, line 35:
```suggestion
// Updates vertices and calculates their displacement.
// Input: newVertices and oldVertices
// Output: verticesDisplacement and newVertices
__global__ void kUpdateVertices(size_t vertexCount, Vec3f* newVerticesToDisplacement, Vec3f* oldToNewVertices)
```
1. Such a comment clarifies what is going on here (no need to refer to Mesh::updateVertices and wasting time)
2. Renaming `newVerticesToVelocity` to `newVerticesToDisplacement`

src/scene/Entity.cpp
```diff
@@ -27,8 +27,10 @@ Entity::Entity(std::shared_ptr<Mesh> mesh) : mesh(std::move(mesh)) {}
 
 void Entity::setTransform(Mat3x4f newTransform)
 {
-	transform = newTransform;
-	Scene::instance().requestASRebuild();
+	formerTransform = transform;
+	transform = {newTransform, Scene::instance().getTime()};
```
msz-rai, line 31: 1. What if the plugin updates transforms first, then sets time on the scene (or mixes up these operations depending on callbacks order)? I'd specify our requirements in the documentation (`core.h` file?)
2. I am considering renaming `transform` into `transformInfo` because it is no longer just transform.

msz-rai (approved):
LGTM!
Two comments to consider

msz-rai:
Also, I found that `Scene::clear` doesn't clear `time` (and `prevTime`). With this feature, it may be a crucial bug when restarting a simulation.

src/scene/Entity.cpp
prybicki: Done.

PR #220, opened 2023-11-17
Fix nodes validation on rays change

msz-rai:


msz-rai:
@prybicki 
Could `GraphRunCtx->makeDirty()` be also used in add/remove child instead of `detachAndDestroy()`? E.g. [here](https://github.com/RobotecAI/RobotecGPULidar/blob/main/src/graph/Node.cpp#L39)

src/graph/GraphRunCtx.hpp
```diff
@@ -63,6 +63,16 @@ struct GraphRunCtx
...
+	/**
+	 * Makes all nodes dirty.
+	 */
+	void makeDirty()
+	{
+		for (auto&& node : nodes) {
+			node->dirty = true;
+		}
+	}
+
```
prybicki, line 75: I'd name it `markNodesDirty`

prybicki (approved):

PR #229, opened 2023-12-06
Calculate velocity derived fields

prybicki:
Bug in optixProgram.cu!


msz-rai:
Bug in optixProgram.cu fixed

test/src/scene/entityVelocityTest.cpp
```diff
@@ -9,6 +9,122 @@
 
 using namespace std::chrono_literals;
 
+constexpr float RadToDeg = (180.0f / M_PI);
```
PawelLiberadzki, line 12: This may be:
std::numbers::pi_v<float>

PawelLiberadzki (approved):

PR #232, opened 2023-12-13
Optimize Tape YAML structure

nebraszka:


src/tape/TapeCall.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Proxy structure for reading TapeCall fields.
+ * Expects a node of the following format:
+ * fnName: {t: <int64_t>, a: [<arg0>, <arg1>, ...]}
+ */
+struct TapeCall
+{
+	std::string getFnName() const { return node.begin()->first.as<std::string>(); }
+	int64_t getTimestampNs() const { return node.begin()->second["t"].as<int64_t>(); }
```
msz-rai, line 27:
```suggestion
	Time getTimestamp() const { return Time::nanoseconds(node.begin()->second["t"].as<std::chrono::nanoseconds::rep>()); }
```
1. I'd use our `Time` class to enable easy conversions between units.
2. Instead of hardcoding the type, I suggest using the exact type of encoded timestamp.

src/tape/TapePlayer.cpp
```diff
@@ -30,20 +31,21 @@ TapePlayer::TapePlayer(const char* path)
...
 
 	yamlRoot = YAML::LoadFile(pathYaml);
-	yamlRecording = yamlRoot["recording"];
 
-	if (yamlRoot[RGL_VERSION]["major"].as<int>() != RGL_VERSION_MAJOR ||
-	    yamlRoot[RGL_VERSION]["minor"].as<int>() != RGL_VERSION_MINOR) {
-		throw RecordError("recording version does not match rgl version");
+	auto version = getTapeCall(findFirst({RGL_VERSION}).value()).getArgsNode();
+	int versionLinear = 1'000'000 * version[0].as<int>() + 1'000 * version[1].as<int>() + version[2].as<int>();
+	int lastTapeUpdate = 16 * 1'000 + 3;
```
msz-rai, line 37: I'd consider moving it to `tapeDefinitions.hpp`:
```
#define RGL_TAPE_FORMAT_VERSION_MAJOR 0
#define RGL_TAPE_FORMAT_VERSION_MINOR 16
#define RGL_TAPE_FORMAT_VERSION_PATCH 3
```

src/tape/TapePlayer.cpp
```diff
@@ -30,20 +31,21 @@ TapePlayer::TapePlayer(const char* path)
...
+	if (versionLinear < lastTapeUpdate) {
+		throw RecordError("Unsupported (outdated) Tape format");
```
msz-rai, line 39: It would be nice to inform which version of tape format is read and which version is required.

src/tape/TapePlayer.cpp
```diff
@@ -30,20 +31,21 @@ TapePlayer::TapePlayer(const char* path)
 	playbackState = std::make_unique<PlaybackState>(pathBin.c_str());
 
 	yamlRoot = YAML::LoadFile(pathYaml);
-	yamlRecording = yamlRoot["recording"];
 
-	if (yamlRoot[RGL_VERSION]["major"].as<int>() != RGL_VERSION_MAJOR ||
-	    yamlRoot[RGL_VERSION]["minor"].as<int>() != RGL_VERSION_MINOR) {
-		throw RecordError("recording version does not match rgl version");
+	auto version = getTapeCall(findFirst({RGL_VERSION}).value()).getArgsNode();
```
msz-rai, line 35: I believe it may fail (pre-refactor tape format version or broken tape). Please add handling for it.

src/tape/TapeRecorder.hpp
```diff
@@ -44,18 +44,64 @@
...
+	uintptr_t valueToYaml(rgl_texture_t value) { return (uintptr_t) value; }
+	uintptr_t valueToYaml(rgl_scene_t* value) { return (uintptr_t) *value; }
+	uintptr_t valueToYaml(rgl_entity_t* value) { return (uintptr_t) *value; }
+	uintptr_t valueToYaml(rgl_texture_t* value) { return (uintptr_t) *value; }
+	size_t valueToYaml(const rgl_vec3f* value) { return writeToBin(value, 1); }
+	size_t valueToYaml(const rgl_mat3x4f* value) { return writeToBin(value, 1); }
+	template<typename T> std::enable_if_t<!std::is_enum_v<T>, T> valueToYaml(T value) { return value; }
+	template<typename T, typename N> size_t valueToYaml(std::pair<T, N> value) { return writeToBin(value.first, value.second); }
+	template<typename N> size_t valueToYaml(std::pair<const void*, N> value) { return writeToBin(static_cast<const char*>(value.first), value.second); }
+	template<typename T> std::enable_if_t<std::is_enum_v<T>, std::string> valueToYaml(T value) { return std::to_string(static_cast<std::underlying_type_t<T>>(value)); }
```
msz-rai, line 102: IMO, the previous formatting style for templated methods was more readable.

msz-rai (approved):
Great refactor :+1: :1st_place_medal: 
I have left some suggestions to consider.

src/tape/TapePlayer.cpp
prybicki: Yes, it should be sufficient to pull out `findFirst({RGL_VERSION})` and check if it `.has_value()`. If not, then report an error.

prybicki (approved):

PR #233, opened 2023-12-15
Add radar postprocessing node

msz-rai:


src/graph/RadarPostprocessPointsNode.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+		if (!isPointClustered) {
+			// Create a new cluster
+			clusters.emplace_back(i, distance, azimuth);
+		}
+	}
+
+	// Merge clusters if are close enough
+	bool allClustersGood = false;
+	while (clusters.size() < 2 || !allClustersGood) {
```
PawelLiberadzki, line 73:
```suggestion
	while (clusters.size() >= 2 && !allClustersGood) {
```
Was this maybe your intention? With above, it will break, when only one cluster left or all clusters are good.

src/graph/RadarPostprocessPointsNode.cpp
```diff
@@ -0,0 +1,195 @@
...
+	bool isDistanceGood = std::abs(minMaxDistance[0] - other.minMaxDistance[1]) <= distanceSeparation &&
+	                      std::abs(minMaxDistance[1] - other.minMaxDistance[0]) <= distanceSeparation;
+
+	bool isAzimuthGood = std::abs(minMaxAzimuth[0] - other.minMaxAzimuth[1]) <= azimuthSeparation &&
+	                     std::abs(minMaxAzimuth[1] - other.minMaxAzimuth[0]) <= azimuthSeparation;
+
+	return isDistanceGood && isAzimuthGood;
+}
+
+void RadarPostprocessPointsNode::RadarCluster::mergeWith(RadarCluster other)
```
PawelLiberadzki, line 184: I would consider passing other as const& - you move indices below but you copy it here either way. Then you should go with e.g. std::transform instead std::move in line 194.

I would also consider mergeIndicesFrom name instead of mergeWith.

PawelLiberadzki (approved):

src/graph/NodesCore.hpp
```diff
@@ -457,3 +457,59 @@ struct GaussianNoiseDistanceNode : IPointsNodeSingleInput
...
+	float azimuthSeparation;
+
+	// RGL related members
+	std::mutex getFieldDataMutex;
+	mutable CacheManager<rgl_field_t, IAnyArray::Ptr> cacheManager;
+
+	struct RadarCluster
+	{
+		RadarCluster(Field<RAY_IDX_U32>::type index, float distance, float azimuth, float elevation);
+		RadarCluster(RadarCluster&& other) = default;
```
PawelLiberadzki, line 499: I believe move constructor (and move assignment operator) has to be marked noexcept to work properly in this case.

include/rgl/api/core.h
```diff
@@ -636,6 +637,19 @@ RGL_API rgl_status_t rgl_node_points_temporal_merge(rgl_node_t* node, const rgl_
...
+/**
+ * Creates or modifies RadarPostprocessPointsNode.
+ * The Node processes point cloud to create radar-like output.
+ * The point cloud is reduced by clustering input based on hit-point distance and hit-point azimuth.
+ * The output consists of the collection of one point per cluster (the closest to the azimuth and elevation center).
+ * Graph input: point cloud
+ * Graph output: point cloud
+ * @param node If (*node) == nullptr, a new Node will be created. Otherwise, (*node) will be modified.
+ * @param distance_separation The maximum distance difference to create a new radar cluster.
+ * @param azimuth_separation The maximum azimuth difference to create a new radar cluster.
```
PawelLiberadzki, line 649: You can consider adding info about units here and for distance separation - may be simpler to use.

PawelLiberadzki (approved):

PR #245, opened 2024-01-22
Implement automatic build of (some) ros2 deps - radar_msgs

prybicki:


extensions/ros2/src/graph/NodesRos2.hpp
```diff
@@ -57,10 +58,7 @@ struct Ros2PublishPointVelocityMarkersNode : IPointsNodeSingleInput
 	using Ptr = std::shared_ptr<Ros2PublishPointVelocityMarkersNode>;
 
 	void setParameters(const char* topicName, const char* frameId, rgl_field_t velocityField);
-	std::vector<rgl_field_t> getRequiredFieldList() const override
-	{
-		return {RGL_FIELD_XYZ_VEC3_F32, RGL_FIELD_RELATIVE_VELOCITY_VEC3_F32};
-	}
+	std::vector<rgl_field_t> getRequiredFieldList() const override { return {XYZ_VEC3_F32, velocityField}; }
```
prybicki, line 61: Smuggling minor fix - state actual required field based on parameters.

extensions/ros2/src/graph/NodesRos2.hpp
```diff
@@ -78,4 +76,24 @@ struct Ros2PublishPointVelocityMarkersNode : IPointsNodeSingleInput
...
+		return {DISTANCE_F32, AZIMUTH_F32, ELEVATION_F32, RADIAL_SPEED_F32};
+	}
+	void validateImpl() override {}
+	void enqueueExecImpl() override {}
+
+private:
+	radar_msgs::msg::RadarScan ros2Message;
+	rclcpp::Publisher<radar_msgs::msg::RadarScan>::SharedPtr ros2Publisher;
+	std::shared_ptr<Ros2InitGuard> ros2InitGuard;
+};
```
prybicki, line 99: Just a stub to ensure it compiles.

setup.py
```diff
@@ -14,7 +15,10 @@ class Config:
     CUDA_MIN_VER_MINOR = 7
     CUDA_MIN_VER_PATCH = 0
     CMAKE_GENERATOR = "'Unix Makefiles'"
-    VCPKG_INSTALL_DIR = os.path.join("external", "vcpkg")
+    RADAR_MSGS_DIR = os.path.join("external", "radar_msgs")
+    RADAR_MSGS_INSTALL_DIR = os.path.join("external", "radar_msgs", "install")
+    RADAR_MSGS_COMMIT = "47d2f26906ef38fa15ada352aea6b5aad547781d"
+    VCPKG_DIR = os.path.join("external", "vcpkg")
```
prybicki, line 21: Renamed to avoid semantic confusion other install dirs. *_DIR refers to location where package is downloaded, *_INSTALL_DIR refers to location where packages has been built, if relevant (install/ dir for ros2 packages)

setup.py
```diff
@@ -66,69 +73,62 @@ def main():
                             help="Pass arguments to ninja. Usage: --ninja=\"args...\". Defaults to \"-j <cpu count>\"")
     args = parser.parse_args()
 
+    # ROS2 standalone obviously implies ROS2
+    if args.with_ros2_standalone:
+        args.with_ros2 = True
```
prybicki, line 78: Moved from below to have semantic correctness just after argument parsing.

setup.py
```diff
@@ -66,69 +73,62 @@ def main():
...
+
+    # Go to script directory
+    os.chdir(sys.path[0])
+
+    # Prepare build directory
+    if args.clean_build and os.path.isdir(args.build_dir):
+        shutil.rmtree(args.build_dir, ignore_errors=True)
```
prybicki, line 85: `radar_msgs` could be removed here, but wasn't sure if that's the intended behavior - would be inconsistent with not removing vcpkg-originating packages.

.gitignore
```diff
@@ -39,6 +39,6 @@ Testing/*
 cmake-build-debug
 cmake-build-release
 build
-external/vcpkg
+external/*
```
msz-rai, line 42: This way, `external/CMakeLists.txt` would be ignored. Please try to add an exception to the rule (and test it):
```suggestion
external/*
!external/CMakeLists.txt
```

extensions/ros2/CMakeLists.txt
```diff
@@ -17,15 +18,21 @@ target_include_directories(RobotecGPULidar PUBLIC
...
     ${visualization_msgs_INCLUDE_DIRS}
+    ${radar_msgs_INCLUDE_DIRS}
 )
 
+# These libraries list, imported from ROS2 packages,
+# contain a large number of duplicates, which results
+# in linking error (LNK1170) due to exceeded line length in rsp file.
+list(REMOVE_DUPLICATES radar_msgs_LIBRARIES)
+list(REMOVE_DUPLICATES visualization_msgs_LIBRARIES)
+list(REMOVE_DUPLICATES sensor_msgs_LIBRARIES)
```
msz-rai, line 29: :1st_place_medal:

setup.py
msz-rai: IMO, it wouldn't be an intended behavior. Dependencies are installed once per machine and there is no need to reinstall them (unless there is a version update - rare case). `--clean-build` as described in `argparse` only removes the build directory before cmake.

setup.py
```diff
@@ -191,5 +193,99 @@ def run_system_command(command: str):
...
+    # Clone radar msgs
+    if not os.path.isdir(cfg.RADAR_MSGS_DIR):
+        run_subprocess_command(
+            f"git clone --single-branch --depth 1 https://github.com/ros-perception/radar_msgs.git {cfg.RADAR_MSGS_DIR}")
+        run_subprocess_command(f"cd {cfg.RADAR_MSGS_DIR} && git checkout {cfg.RADAR_MSGS_COMMIT} && cd ..")
+    # Build radar msgs
+    if not os.path.isdir(cfg.RADAR_MSGS_INSTALL_DIR):
+        original_path = Path.cwd()
+        os.chdir(cfg.RADAR_MSGS_DIR)
+        run_subprocess_command(f"colcon build")
```
msz-rai, line 255: The command fails if ROS2 is not sourced. I'd add a check (the same as `has_colcon()`) and inform the user to source ROS2 before if he didn't do it.

setup.py
```diff
@@ -66,69 +73,62 @@ def main():
...
 
     # Extend Path with libRobotecGPULidar location to link tests properly during the build on Windows
     if on_windows():
         os.environ["Path"] = os.environ["Path"] + ";" + os.path.join(os.getcwd(), args.build_dir)
 
+    if args.with_ros2:
+        # Source environment for additional packages
+        # ROS2 itself must be sourced by the user, because its location is unknown to this script
+        setup = "setup.bat" if on_windows() else "setup.sh"
+        source_environment(os.path.join(os.getcwd(), cfg.RADAR_MSGS_INSTALL_DIR, setup))
```
msz-rai, line 127: I'd add a check if ROS2 is sourced at the beginning of this if-statement. It could produce a nicer output than this:
image

setup.py
```diff
@@ -42,6 +47,8 @@ def main():
                         help="Path to build directory. Default: 'build'")
     parser.add_argument("--install-pcl-deps", action='store_true',
                         help="Install dependencies for PCL extension and exit")
+    parser.add_argument("--install-ros2-deps", action='store_true',
+                        help="Install dependencies for ROS2 extension and exit")
```
msz-rai, line 51: Please update the documentation [here](https://github.com/RobotecAI/RobotecGPULidar/blob/develop/docs/Ros2Extension.md).

msz-rai (commented):
Nice job!
Please address the comments.
~~I am still testing in docker - please wait for the approval.~~
_Edit: I cannot test it in docker without changes from this [PR](https://github.com/RobotecAI/RobotecGPULidar/pull/240)_

.gitignore
prybicki: Great catch! Tested your suggestion and it works.

extensions/ros2/CMakeLists.txt
prybicki: :bow:

setup.py
prybicki: Done

setup.py
msz-rai: Seems to not be addressed yet.

msz-rai (approved):
LGTM

PR #253, opened 2024-02-13
Fix OptiX crash on Windows

prybicki:


src/Optix.cpp
```diff
@@ -52,8 +52,6 @@ static CUcontext getCurrentDeviceContext()
 	const char* error = nullptr;
 	CUresult status;
 
-	cudaFree(nullptr); // Force CUDA runtime initialization
-
```
prybicki, line 56: Removed, because CUDA runtime is actually initialized earlier.

PawelLiberadzki (approved):

PR #257, opened 2024-02-26
Refactor raytrace node configurations and introduce non-hit distance handling

nebraszka:


src/gpu/optixPrograms.cu
```diff
@@ -108,6 +108,14 @@ __forceinline__ __device__ void saveRayResult(const Vec3f& xyz = Vec3f{NON_HIT_V
...
 {
 	if (ctx.scene == 0) {
+		if (ctx.doCustomNonHitDistances) {
+			Mat3x4f ray = ctx.rays[optixGetLaunchIndex().x];
+			Vec3f origin = ray * Vec3f{0, 0, 0};
+			Vec3f dir = ray * Vec3f{0, 0, 1} - origin;
+			Vec3f xyz = origin + dir.normalized() * ctx.farNonHitDistance;
+			saveRayResult<false>(xyz, ctx.farNonHitDistance);
+			return;
+		}
```
msz-rai, line 118: I would think about reducing repetitive code here in `optixPrograms.cu`. We can huddle to discuss it.

test/src/TapeTest.cpp
```diff
@@ -207,7 +207,7 @@ TEST_F(TapeTest, RecordPlayAllCalls)
 	EXPECT_RGL_SUCCESS(rgl_node_rays_set_ring_ids(&setRingIds, rings.data(), rings.size()));
 
 	rgl_node_t setTimeOffsets = nullptr;
-	std::vector<float> timeOffsets = {0.0f, 1.0f, 2.0f};
+	std::vector<float> timeOffsets = {1.0f, 2.0f};
```
msz-rai, line 210: I am curious about this change :smiley:

test/src/TapeTest.cpp
```diff
@@ -223,13 +223,11 @@ TEST_F(TapeTest, RecordPlayAllCalls)
...
 
-	rgl_node_t raytraceWithDistortion = nullptr;
 	rgl_vec3f linearVelocity{1.0f, 2.0f, 3.0f};
 	rgl_vec3f angularVelocity{1.0f, 2.0f, 3.0f};
-	EXPECT_RGL_SUCCESS(rgl_node_raytrace_with_distortion(&raytraceWithDistortion, nullptr, &linearVelocity, &angularVelocity));
+	EXPECT_RGL_SUCCESS(rgl_node_raytrace_configure_velocity(raytrace, &linearVelocity, &angularVelocity));
 
-	rgl_node_t raytraceInMotion = nullptr;
-	EXPECT_RGL_SUCCESS(rgl_node_raytrace_in_motion(&raytraceInMotion, nullptr, &linearVelocity, &angularVelocity, false));
+	EXPECT_RGL_SUCCESS(rgl_node_raytrace_configure_distortion(raytrace, true));
```
msz-rai, line 230: `rgl_node_raytrace_configure_non_hit_distance_values` is missing here.

test/src/graph/nodes/RaytraceNodeTest.cpp
```diff
@@ -21,3 +29,200 @@ TEST_F(RaytraceNodeTest, valid_arguments)
...
+	                            "angular_velocity != nullptr");
+}
+
+TEST_F(RaytraceNodeTest, config_velocity_invalid_node_object)
+{
+	const rgl_vec3f linearVelocity = {1.0f, 2.0f, 3.0f};
+	const rgl_vec3f angularVelocity = {4.0f, 5.0f, 6.0f};
+	EXPECT_RGL_INVALID_OBJECT(rgl_node_raytrace_configure_velocity((rgl_node_t) 0x1234, &linearVelocity, &angularVelocity),
+	                          "Object does not exist: Node 0x1234");
+}
```
msz-rai, line 62: IMO, it could be wrapped into one test `config_velocity_invalid_arguments`

test/src/graph/nodes/RaytraceNodeTest.cpp
```diff
@@ -21,3 +29,200 @@ TEST_F(RaytraceNodeTest, valid_arguments)
...
+TEST_F(RaytraceNodeTest, config_distortion_invalid_argument_node)
+{
+	EXPECT_RGL_INVALID_ARGUMENT(rgl_node_raytrace_configure_distortion(raytraceNode, false), "node != nullptr");
+}
+
+TEST_F(RaytraceNodeTest, config_distortion_invalid_node_object)
+{
+	EXPECT_RGL_INVALID_OBJECT(rgl_node_raytrace_configure_distortion((rgl_node_t) 0x1234, false),
+	                          "Object does not exist: Node 0x1234");
+}
```
msz-rai, line 82: ditto

test/src/graph/nodes/RaytraceNodeTest.cpp
```diff
@@ -21,3 +29,200 @@ TEST_F(RaytraceNodeTest, valid_arguments)
...
+{
+	EXPECT_RGL_INVALID_ARGUMENT(rgl_node_raytrace_configure_non_hit_distance_values(raytraceNode, 0.0f, 0.0f),
+	                            "node != nullptr");
+}
+
+TEST_F(RaytraceNodeTest, config_non_hit_distance_invalid_node_object)
+{
+	EXPECT_RGL_INVALID_OBJECT(rgl_node_raytrace_configure_non_hit_distance_values((rgl_node_t) 0x1234, 0.0f, 0.0f),
+	                          "Object does not exist: Node 0x1234");
+}
```
msz-rai, line 101: ditto

test/src/graph/nodes/RaytraceNodeTest.cpp
```diff
@@ -21,3 +29,200 @@ TEST_F(RaytraceNodeTest, valid_arguments)
...
+
+TEST_F(RaytraceNodeTest, config_non_hit_distance_valid_arguments)
+{
+	ASSERT_RGL_SUCCESS(rgl_node_raytrace(&raytraceNode, nullptr));
+
+	EXPECT_RGL_SUCCESS(rgl_node_raytrace_configure_non_hit_distance_values(raytraceNode, 0.0f, 0.0f));
+}
+
+TEST_F(RaytraceNodeTest, config_non_hit_near_distance_should_correctly_change_output)
```
msz-rai, line 110: I'd set up the scene for both hits and non-hits to be produced. The test will validate whether hits are not changed.
The same goes for the second test.

msz-rai (changes requested):
Good job :+1: 
The proposed changes make `NON_HIT_VALUE` to be inconsistent in some scenarios. It needs to be corrected (I have left a suggestion for it).
The rest of the comments are more about cosmetic changes.

msz-rai:
\+ Please rebase PR onto the latest `develop` branch to activate CI :wink:

test/src/TapeTest.cpp
nebraszka: In an earlier version of this test, both raytraceWithDistortion node and setTimeOffsets node were not included in the graph. When I attached them, I found that the test fails because the number of time offsets does not match the number of rays. Hence the change - we have two rays, so we want two time offsets.

src/gpu/optixPrograms.cu
```diff
@@ -105,10 +103,20 @@ __forceinline__ __device__ void saveRayResult(const Vec3f& xyz = Vec3f{NON_HIT_V
 	}
 }
 
+__forceinline__ __device__ void saveNonHitRayResult(float nonHitDistance)
+{
+	Mat3x4f ray = ctx.rays[optixGetLaunchIndex().x];
+	Vec3f origin = ray * Vec3f{0, 0, 0};
+	Vec3f dir = ray * Vec3f{0, 0, 1} - origin;
+	Vec3f xyz = origin + dir.normalized() * nonHitDistance;
+	xyz = {isnan(xyz.x()) ? 0 : xyz.x(), isnan(xyz.y()) ? 0 : xyz.y(), isnan(xyz.z()) ? 0 : xyz.z()};
```
msz-rai, line 112:
```suggestion
	Vec3f displacement = dir.normalized() * nonHitDistance;
	displacement = {isnan(displacement.x()) ? 0 : displacement.x(), isnan(displacement.y()) ? 0 : displacement.y(),
	                isnan(displacement.z()) ? 0 : displacement.z()};
	Vec3f xyz = origin + displacement;
```
Displacement is zero, not XYZ.

test/include/helpers/commonHelpers.hpp
```diff
@@ -10,6 +10,7 @@
 
 static constexpr float EPSILON_F = 1e-6f;
 static constexpr int maxGPUCoresTestCount = 20000;
+static constexpr float INF = std::numeric_limits<float>::infinity();
```
msz-rai, line 13: I'd avoid polluting header files with such aliases (especially when it is used in `DistanceFieldTest.cpp` only). I'd move it there.

test/src/graph/VelocityDistortionTest.cpp
```diff
@@ -43,38 +45,52 @@ TEST_F(VelocityDistortionTest, smoke_test)
...
+	std::vector<rgl_field_t> outFields{XYZ_VEC3_F32, IS_HIT_I32};
+
 	// Create nodes
 	EXPECT_RGL_SUCCESS(rgl_node_rays_from_mat3x4f(&useRays, rays.data(), rayCount));
 	EXPECT_RGL_SUCCESS(rgl_node_rays_set_time_offsets(&offsetsNode, timeOffsets.data(), rayCount));
-	EXPECT_RGL_SUCCESS(rgl_node_raytrace_with_distortion(&raytrace, nullptr, &linearVelocity, &angularVelocity));
+	EXPECT_RGL_SUCCESS(rgl_node_raytrace(&raytrace, nullptr));
+	EXPECT_RGL_SUCCESS(rgl_node_raytrace_configure_velocity(raytrace, &linearVelocity, &angularVelocity));
+	EXPECT_RGL_SUCCESS(rgl_node_raytrace_configure_distortion(raytrace, true));
+	EXPECT_RGL_SUCCESS(rgl_node_raytrace_configure_non_hits(raytrace, INF, INF));
```
msz-rai, line 56: There is no need to configure non-hits here. Also, the configuration is the same as the default one.

test/src/graph/nodes/RaytraceNodeTest.cpp
```diff
@@ -21,3 +29,249 @@ TEST_F(RaytraceNodeTest, valid_arguments)
...
+	/**
+	 * Hits with raytrace non hits configuration scenario
+	 */
+
+	rgl_node_t rangeRaysNode = nullptr;
+	std::vector<rgl_vec2f> ranges{
+	    rgl_vec2f{0.0f, CUBE_HALF_EDGE + 1.0f}  // Ray should hit the cube
+	};
+
+	ASSERT_RGL_SUCCESS(rgl_node_rays_set_range(&rangeRaysNode, ranges.data(), ranges.size()));
```
msz-rai, line 151: Why this additional node is added in the second scenario? I would expect only a change in the non-hits configuration. Now, the test validates not only non-hits configuration but `rgl_node_rays_set_range` also. I would leave the pipeline and the same very the same between runs.

include/rgl/api/core.h
```diff
@@ -558,53 +558,41 @@ RGL_API rgl_status_t rgl_node_points_transform(rgl_node_t* node, const rgl_mat3x
...
+ *  - for non-hits closer than a minimum range (`near`),
+ *  - for non-hits beyond a maximum range (`far`).
+ * Concurrently, it computes the RGL_FIELD_XYZ_VEC3_F32 field for these non-hit scenarios based on these distances, along with ray origin and direction.
+ * @param node RaytraceNode to modify.
+ * @param near Distance value for non-hits closer than minimum range.
+ * @param far Distance value for non-hits beyond maximum range.
  */
-RGL_API rgl_status_t rgl_node_raytrace_in_motion(rgl_node_t* node, rgl_scene_t scene, const rgl_vec3f* linear_velocity,
-                                                 const rgl_vec3f* angular_velocity, bool apply_ray_distortion);
+RGL_API rgl_status_t rgl_node_raytrace_configure_non_hits(rgl_node_t node, float near, float far);
```
msz-rai, line 595: Please add info about the default distance for non-hits.

msz-rai (changes requested):

msz-rai (approved):
:+1: could be merged if CI passes :robot:

PR #262, opened 2024-03-11
Add is_valid functions to entity, mesh and texture in public API

AlgoryxJosef:
This is really useful for example where some application or library has reference-counted wrappers of rgl-entities, rgl-meshes or rgl-textures which calls rgl_xxx_destroy once the last referenced object is destroyed. If some lingering reference exists (for example in client code) after e.g. rgl_cleanup() has been called called, an error is produced since then, rgl_xxx_destroy is called after rgl_cleanup. 

So by adding these functions, the application or library can check if the entity/mesh/texture is still valid, and only if it is, call rgl_xxx_destroy. For example. There probably exists other use cases as well.

### DIFF
--- a/include/rgl/api/core.h
+++ b/include/rgl/api/core.h
@@ -460,6 +460,14 @@ RGL_API rgl_status_t rgl_mesh_destroy(rgl_mesh_t mesh);
  */
 RGL_API rgl_status_t rgl_mesh_update_vertices(rgl_mesh_t mesh, const rgl_vec3f* vertices, int32_t vertex_count);
 
+/**
+ * Assigns value true to out_valid if the given mesh is known and has not been destroyed,
+ * assigns value false otherwise.
+ * @param mesh Mesh to check for validity
+ * @param out_valid Boolean set to indicate validity
+ */
+RGL_API rgl_status_t rgl_is_mesh_valid(rgl_mesh_t mesh, bool* out_valid);
+
 /******************************** ENTITY ********************************/
 
 /**
@@ -496,9 +504,17 @@ RGL_API rgl_status_t rgl_entity_set_id(rgl_entity_t entity, int32_t id);
 /**
  * Assign intensity texture to the given Entity. The assumption is that the Entity can hold only one intensity texture.
  * @param entity Entity to modify.
- * @apram texture Texture to assign.
+ * @param texture Texture to assign.
  */
 RGL_API rgl_status_t rgl_entity_set_intensity_texture(rgl_entity_t entity, rgl_texture_t texture);
+
+/**
+ * Assigns value true to out_valid if the given entity is known and has not been destroyed,
+ * assigns value false otherwise.
+ * @param entity Entity to check for validity
+ * @param out_valid Boolean set to indicate validity
+ */
+RGL_API rgl_status_t rgl_is_entity_valid(rgl_entity_t entity, bool* out_valid);
 
 /******************************* TEXTURE *******************************/
 
@@ -515,9 +531,17 @@ RGL_API rgl_status_t rgl_texture_create(rgl_texture_t* out_texture, const void* 
 /**
  * Informs that the given texture will be no longer used.
  * The texture will be destroyed after all referring Entities are destroyed.
- * @param mesh Texture to be marked as no longer needed
+ * @param texture Texture to be marked as no longer needed
  */
 RGL_API rgl_status_t rgl_texture_destroy(rgl_texture_t texture);
+
+/**
+ * Assigns value true to out_valid if the given texture is known and has not been destroyed,
+ * assigns value false otherwise.
+ * @param texture Texture to check for validity
+ * @param out_valid Boolean set to indicate validity
+ */
+RGL_API rgl_status_t rgl_is_texture_valid(rgl_texture_t texture, bool* out_valid);
 
 /******************************** SCENE ********************************/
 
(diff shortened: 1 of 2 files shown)


src/api/apiCore.cpp
```diff
@@ -256,6 +256,15 @@ void TapeCore::tape_mesh_update_vertices(const YAML::Node& yamlNode, PlaybackSta
 	                         yamlNode[2].as<int32_t>());
 }
 
+rgl_status_t rgl_is_mesh_valid(rgl_mesh_t mesh, bool* out_valid)
+{
+	auto status = rglSafeCall([&]() {
+		*out_valid = Mesh::instances.find(mesh) != Mesh::instances.end();
+	});
+	TAPE_HOOK();
+	return status;
```
prybicki, line 265: - Added `CHECK_ARG` to ensure RGL fails in a controllable manner if the incoming pointer is null.
- Passed API call arguments to `TAPE_HOOK`. Tape is our home-baked solution to record all API calls to a file and then replay them from a file. It was created to eliminate the need to run the original application (often being a heavy game engine) when debugging RGL.
- Since we're using C++20, we can use `std::unorderd_map::contains(...)`
```suggestion
	auto status = rglSafeCall([&]() {
		CHECK_ARG(out_valid != nullptr);
		*out_valid = Mesh::instances.contains(mesh);
	});
	TAPE_HOOK(mesh, out_valid);
	return status;
```

src/api/apiCore.cpp
```diff
@@ -355,6 +364,15 @@ void TapeCore::tape_entity_set_intensity_texture(const YAML::Node& yamlNode, Pla
 	                                 state.textures.at(yamlNode[1].as<TapeAPIObjectID>()));
 }
 
+rgl_status_t rgl_is_entity_valid(rgl_entity_t entity, bool* out_valid)
+{
+	auto status = rglSafeCall([&]() {
+		*out_valid = Entity::instances.find(entity) != Entity::instances.end();
+	});
+	TAPE_HOOK();
+	return status;
```
prybicki, line 373:
```suggestion
	auto status = rglSafeCall([&]() {
		CHECK_ARG(out_valid != nullptr);
		*out_valid = Entity::instances.contains(entity);
	});
	TAPE_HOOK(entity, out_valid);
	return status;
```

src/api/apiCore.cpp
```diff
@@ -398,6 +416,15 @@ void TapeCore::tape_texture_destroy(const YAML::Node& yamlNode, PlaybackState& s
...
 }
 
+rgl_status_t rgl_is_texture_valid(rgl_texture_t texture, bool* out_valid)
+{
+	auto status = rglSafeCall([&]() {
+		*out_valid = Texture::instances.find(texture) != Texture::instances.end();
+	});
+	TAPE_HOOK();
+	return status;
+}
```
prybicki, line 426:
```suggestion
rgl_status_t rgl_is_texture_valid(rgl_texture_t texture, bool* out_valid)
{
	auto status = rglSafeCall([&]() {
		CHECK_ARG(out_valid != nullptr);
		*out_valid = Texture::instances.contains(texture);
	});
	TAPE_HOOK(texture, out_valid);
	return status;
}
```

prybicki (changes requested):
I understand the use-case you have described and I think the PR is a very reasonable proposition. Thank you for your time and contribution!

I've added a few suggestions to improve the code to follow our conventions. Also, for the sake of API completeness, please add an `is_alive` API call for the `rgl_node_t` too.

include/rgl/api/core.h
```diff
@@ -460,6 +460,14 @@ RGL_API rgl_status_t rgl_mesh_destroy(rgl_mesh_t mesh);
  */
 RGL_API rgl_status_t rgl_mesh_update_vertices(rgl_mesh_t mesh, const rgl_vec3f* vertices, int32_t vertex_count);
 
+/**
+ * Assigns value true to out_valid if the given mesh is known and has not been destroyed,
+ * assigns value false otherwise.
+ * @param mesh Mesh to check for validity
+ * @param out_valid Boolean set to indicate validity
+ */
+RGL_API rgl_status_t rgl_is_mesh_valid(rgl_mesh_t mesh, bool* out_valid);
```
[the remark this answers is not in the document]
AlgoryxJosef: Thank you for the pleasent and helful responses.

Renamed to is_alive fixed in: [a99dd3b](https://github.com/RobotecAI/RobotecGPULidar/pull/262/commits/a99dd3b6ff804c34b508b6aa5cf15bb9f417b9be) 
    
Added is_alive function for Node in: [648d1f3](https://github.com/RobotecAI/RobotecGPULidar/pull/262/commits/648d1f387eb8d700df471c3919261d7d0017dd60)

src/api/apiCore.cpp
AlgoryxJosef: Thank you. This has been fixed, see comment above.

src/api/apiCore.cpp
AlgoryxJosef: Thank you. This has been fixed, see comment above.

AlgoryxJosef:
@prybicki do you have a few minutes to spare looking at this? All discussions should be resolved.

prybicki (approved):

prybicki:
@AlgoryxJosef 

Sure. I resolved your comments, approved PR and I will merge it once it passes our CI.

PR #265, opened 2024-03-13
Fix tape realtime play

prybicki:


src/tape/TapePlayer.cpp
```diff
@@ -23,14 +24,11 @@
...
+TapePlayer::TapePlayer(const char* path) : path(path)
 {
-	std::string pathYaml = fs::path(path).concat(YAML_EXTENSION).string();
-	std::string pathBin = fs::path(path).concat(BIN_EXTENSION).string();
-
-	playbackState = std::make_unique<PlaybackState>(pathBin.c_str());
+	playbackState = std::make_unique<PlaybackState>(getBinPath().c_str());
 
-	yamlRoot = YAML::LoadFile(pathYaml);
+	yamlRoot = YAML::LoadFile(getYamlPath());
```
PiotrMrozik, line 31: `getYamlPath`
I used only there I suppose. Is it worth it to encapsulate it in the method?

PiotrMrozik (approved):

src/tape/TapePlayer.cpp
prybicki: `getBinPath()` is used in more places, so I did it just for the sake of completeness/symmetry.

PR #270, opened 2024-03-22
Fix windows build error due to windows header definition leakage

AlgoryxJosef:
A few functions in the develop branch uses parameters "near" and "far".
On Windows, these are defined in minwindef.h which causes build error(s).

This PR fixes this build error by undefining these keywords if they are defined (Windows only).

### DIFF
--- a/include/rgl/api/core.h
+++ b/include/rgl/api/core.h
@@ -644,14 +644,14 @@ RGL_API rgl_status_t rgl_node_raytrace_configure_distortion(rgl_node_t node, boo
  * Modifies RaytraceNode to set non-hit values for distance.
  * Default non-hit value for the RGL_FIELD_DISTANCE_F32 field is set to infinity.
  * This function allows to set custom values:
- *  - for non-hits closer than a minimum range (`near`),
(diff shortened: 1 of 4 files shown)


msz-rai:
Great catch @AlgoryxJosef! Thank you :+1:

We haven't tested `develop` branch on Windows yet.

I wonder if changing parameter names wouldn't be a better solution. We would avoid code duplication in two .cpp files. I propose the change `near` -> `nearDistance` and `far` -> `farDistance`. What do you think?

AlgoryxJosef:
@msz-rai, thank you for the feedback on this. I suppose the renaming fix is OK. The near/far definition will be left as a mine in the future though on Windows, but perhaps that's not a big issue.

Fixed in: https://github.com/RobotecAI/RobotecGPULidar/pull/270/commits/d340e6e6b40dde4e5eca0bd6e6bfbcac408217eb

msz-rai:
Thank you. After the CI check, I will merge this PR.

msz-rai (approved):

Issue #271, opened 2024-03-22
Build error on Windows - identifier "M_PIf" is undefined


AlgoryxJosef:
Building the develop branch on Windows, Visual Studio 2019. Commit 1cc5becd25585282c685c169f720221c834056b3.

Have tried Cuda Toolkit version 11.4.4 and 11.8.89, as well as Optix 7.2 and 7.6, in all combinations.

Error:  

`C:\Users\Admin\git\fork\RobotecGPULidar\src\gpu\nodeKernels.cu(108): error: identifier "M_PIf" is undefined`
 
I suppose some include is missing, but I could not for the life of me figure out which one. On Linux I guess M_PIf is defined, but not on Windows. 
Have fixed this locally by simply hard-coding 3.141592f, but a real solution would be great.  

Side note: I also hit upon another build error, but managed to fix that one, have created a PR for it: https://github.com/RobotecAI/RobotecGPULidar/pull/270

msz-rai:
Fixed with https://github.com/RobotecAI/RobotecGPULidar/pull/272

PR #272, opened 2024-03-25
Remove M_PIf to enable compilation on Windows

prybicki:


test/src/scene/rcsAngleDistributionTest.cpp
```diff
@@ -24,8 +24,8 @@ const auto minElevation = -7.5f;
 const auto maxElevation = 7.5f;
 const auto azimuthStep = 0.05f;
 const auto elevationStep = 0.05f;
-const auto azimuthRad = (maxAzimuth - minAzimuth) * M_PIf / 180.0f;
-const auto elevationRad = (maxElevation - minElevation) * M_PIf / 180.0f;
+const auto azimuthRad = (maxAzimuth - minAzimuth) * static_cast<float>(M_PI) / 180.0f;
+const auto elevationRad = (maxElevation - minElevation) * static_cast<float>(M_PI) / 180.0f;
```
msz-rai, line 28:
```suggestion
const auto azimuthRad = (maxAzimuth - minAzimuth) * std::numbers::pi_v<float> / 180.0f;
const auto elevationRad = (maxElevation - minElevation) * std::numbers::pi_v<float> / 180.0f;
```
with C++20 we can do it slightly better

msz-rai (approved):
LGTM :+1: 
It should resolve this issue: https://github.com/RobotecAI/RobotecGPULidar/issues/271

PR #273, opened 2024-03-25
Improve rcs

msz-rai:


test/src/scene/radarTest.cpp
```diff
@@ -26,6 +35,80 @@ const auto maxElevation = 7.5f;
 const auto azimuthStep = 0.49f;
 const auto elevationStep = 0.49f;
 
+Mat3x4f calculateMat(float azi, float ele, float rol)
```
PawelLiberadzki, line 38: This function may go out at all - it was a substitute for our rotationDeg function and it does not help in any way.

PawelLiberadzki (approved):
Rest looks fine, considering it is a pull to feature branch.

PR #276, opened 2024-04-15
Feature/fault injection

PiotrMrozik:


msz-rai:
Didn't we agree that the implementation should be on the rays + raytracing side (e.g. a new property for the ray that informs the ray is dirty and raytracing would produce a non-hit for that ray)?

The current implementation modifies the isHit field only, keeping hit values for other fields. It matters because some outputs expect that non-hits have all fields consistent (e.g. with API call [rgl_node_raytrace_configure_non_hits](https://github.com/RobotecAI/RobotecGPULidar/blob/develop/include/rgl/api/core.h#L696) we can configure distance field value for non-hit)

PR #277, opened 2024-04-17
Implement sub-sampling in raytracing code

prybicki:


src/gpu/optixPrograms.cu
```diff
@@ -266,17 +224,96 @@ extern "C" __global__ void __closesthit__()
...
+	return Mat3x4f::rotationRad(0, 0, static_cast<float>(vertexIdx) * vertexAngleStep) * // Second
+	       Mat3x4f::rotationRad(circleDivergence, 0, 0);
+}
 
-extern "C" __global__ void __anyhit__() {}
+__device__ void saveNonHitRayResult(float nonHitDistance)
+{
+	Mat3x4f ray = ctx.rays[optixGetLaunchIndex().x];
+	Vec3f origin = ray * Vec3f{0, 0, 0};
+	Vec3f dir = ray * Vec3f{0, 0, 1} - origin;
```
msz-rai, line 266: Wouldn't be faster by using `optixGetWorldRayDirection()` and `optixGetWorldRayOrigin()`?

src/gpu/optixPrograms.cu
```diff
@@ -266,17 +224,96 @@ extern "C" __global__ void __closesthit__()
...
+	Vec3f displacement = dir.normalized() * nonHitDistance;
+	displacement = {isnan(displacement.x()) ? 0 : displacement.x(), isnan(displacement.y()) ? 0 : displacement.y(),
+	                isnan(displacement.z()) ? 0 : displacement.z()};
+	Vec3f xyz = origin + displacement;
+	saveRayResult<false>(xyz, nonHitDistance, 0, RGL_ENTITY_INVALID_ID, Vec3f{NAN}, Vec3f{NAN}, 0.001f, Vec3f{NAN}, NAN);
```
msz-rai, line 271: `0.001f` :mag:

src/gpu/optixPrograms.cu
```diff
@@ -141,69 +69,91 @@ extern "C" __global__ void __raygen__()
...
-	Vec3f origin = ray * Vec3f{0, 0, 0};
-	Vec3f dir = ray * Vec3f{0, 0, 1} - origin;
 	float maxRange = ctx.rayRangesCount == 1 ? ctx.rayRanges[0].y() : ctx.rayRanges[rayIdx].y();
 
-	unsigned int flags = OPTIX_RAY_FLAG_DISABLE_ANYHIT;
-	Vec3fPayload originPayload = encodePayloadVec3f(origin);
-	optixTrace(ctx.scene, origin, dir, 0.0f, maxRange, 0.0f, OptixVisibilityMask(255), flags, 0, 1, 0, originPayload.p0,
-	           originPayload.p1, originPayload.p2);
+	shootSamplingRay(ray, maxRange, 0); // Shoot primary ray
+	if (ctx.beamHalfDivergence > 0.0f) {
```
msz-rai, line 82: I would make multi-sampling dependent on the expected result or raytrace setting, not on the lidar configuration. To be resolved in the next PR anyway.

src/graph/RaytraceNode.cpp
```diff
@@ -65,6 +65,10 @@ void RaytraceNode::enqueueExecImpl()
 		data->resize(raysNode->getRayCount(), false, false);
 	}
 
+	mrSamples.resize(MULTI_RETURN_BEAM_SAMPLES * raysNode->getRayCount());
+	mrFirst.resize(raysNode->getRayCount());
+	mrLast.resize(raysNode->getRayCount());
```
msz-rai, line 70: No need to be allocated if we do not compute multi-return

msz-rai (approved):
Good job :+1: 
I have left some minor comments to address, but the code looks good :+1:

src/gpu/optixPrograms.cu
prybicki: Why would you do so? The only reason to set `beamDivergence` is when you're interested in multi-return output. This way, `beamHalfDivergence` becomes the only source of truth whether we should perform multi-sampling.

This is different from LiDAR velocity and applying distortion, because LiDAR velocity is used also to compute `RELATIVE_VELOCITY`, so we need additional `ctx.doApplyDistortion` there.

src/gpu/optixPrograms.cu
```diff
@@ -141,69 +69,91 @@ extern "C" __global__ void __raygen__()
...
+				const unsigned beamSampleRayIdx = 1 + circleIdx * MULTI_RETURN_BEAM_VERTICES + vertexIdx;
+				shootSamplingRay(sampleRay, maxRange, beamSampleRayIdx);
+			}
+		}
+	}
+}
+
+extern "C" __global__ void __miss__()
+{
+	const int beamIdx = (int) optixGetLaunchIndex().x;
```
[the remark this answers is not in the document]
prybicki: Applied

src/gpu/optixPrograms.cu
```diff
@@ -141,69 +69,91 @@ extern "C" __global__ void __raygen__()
...
-	Vec3f hitObject = Vec3f((1 - u - v) * A + u * B + v * C);
-	Vec3f hitWorld = optixTransformPointFromObjectToWorldSpace(hitObject);
-
-	int objectID = optixGetInstanceId();
+	// Ray
+	const int beamIdx = optixGetLaunchIndex().x;
+	const unsigned beamSampleRayIdx = optixGetPayload_0();
+	const unsigned circleIdx = (beamSampleRayIdx - 1) / MULTI_RETURN_BEAM_VERTICES;
+	const unsigned vertexIdx = (beamSampleRayIdx - 1) % MULTI_RETURN_BEAM_VERTICES;
+	const unsigned mrSamplesIndex = beamIdx * MULTI_RETURN_BEAM_SAMPLES + beamSampleRayIdx;
```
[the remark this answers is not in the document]
prybicki: Applied manually

src/gpu/optixPrograms.cu
prybicki: It would be faster, but it does not work since this function is called from `__raygen__`, where those functions are illegal to call :oncoming_police_car:

src/gpu/optixPrograms.cu
prybicki: I'm very puzzled seeing this. I've looked into the git blame mirror and I saw myself. No idea why this value. I've changed it to `NAN` and tests are passing.

src/graph/RaytraceNode.cpp
prybicki: Good catch. For now, I cannot check it, but I added a TODO.

src/gpu/optixPrograms.cu
msz-rai: Resolved offline

PR #283, opened 2024-05-06
Add API call to configure beam divergence

prybicki:


include/rgl/api/core.h
```diff
@@ -713,6 +713,15 @@ RGL_API rgl_status_t rgl_node_raytrace_configure_non_hits(rgl_node_t node, float
...
 RGL_API rgl_status_t rgl_node_raytrace_configure_mask(rgl_node_t node, const int8_t* rays_mask, int32_t rays_count);
 
+/**
+ * Modifies RaytraceNode to set beam divergence.
+ * Beam divergence is used to calculate the beam width at the distance of hit point.
+ * Setting beam divergence > 0.0f is required to use query for multi-return results.
+ * @param node RaytraceNode to modify.
+ * @param beamDivergence Beam divergence in radians.
+ */
+RGL_API rgl_status_t rgl_node_raytrace_configure_beam_divergence(rgl_node_t node, float beamDivergence);
```
PawelLiberadzki, line 723:
```suggestion
RGL_API rgl_status_t rgl_node_raytrace_configure_beam_divergence(rgl_node_t node, float beam_divergence);
```
I assume snake case should be used here. And `rgl_node_raytrace_configure_non_hits` just has the same mistake.

include/rgl/api/core.h
```diff
@@ -713,6 +713,15 @@ RGL_API rgl_status_t rgl_node_raytrace_configure_non_hits(rgl_node_t node, float
  */
 RGL_API rgl_status_t rgl_node_raytrace_configure_mask(rgl_node_t node, const int8_t* rays_mask, int32_t rays_count);
 
+/**
+ * Modifies RaytraceNode to set beam divergence.
+ * Beam divergence is used to calculate the beam width at the distance of hit point.
+ * Setting beam divergence > 0.0f is required to use query for multi-return results.
```
PawelLiberadzki, line 719: Can I maybe set beam divergence to 0.0f through this call, in order to disable it? This partially refers to line 949 in apiCore.cpp, where you check for `divergence >= 0.0f`.

PawelLiberadzki (approved):

include/rgl/api/core.h
prybicki: Yes, that was my intention. I'll mention this in the documentation, thanks.

include/rgl/api/core.h
prybicki: Yes, you are correct, thanks for the catch.

PR #285, opened 2024-05-15
Architectural changes to support multi-return feature

prybicki:


src/gpu/nodeKernels.cu
```diff
@@ -229,25 +229,32 @@ __global__ void kProcessBeamSamplesFirstLast(size_t beamCount, int samplesPerBea
...
-		    beamSamples.distance[beamIdx * samplesPerBeam + lastIdx]) {
+		if (beamSamples.distance[beamIdx * samplesPerBeam + sampleIdx] > currentLastDistance) {
 			lastIdx = sampleIdx;
 		}
 	}
-	first.xyz[beamIdx] = beamSamples.xyz[beamIdx * samplesPerBeam + firstIdx];
-	first.distance[beamIdx] = beamSamples.distance[beamIdx * samplesPerBeam + firstIdx];
-	last.xyz[beamIdx] = beamSamples.xyz[beamIdx * samplesPerBeam + lastIdx];
-	last.distance[beamIdx] = beamSamples.distance[beamIdx * samplesPerBeam + lastIdx];
+	first.isHit[beamIdx] = firstIdx > 0;
```
nebraszka, line 247: Shouldn't this condition be >= ?

src/gpu/nodeKernels.cu
```diff
@@ -229,25 +229,32 @@ __global__ void kProcessBeamSamplesFirstLast(size_t beamCount, int samplesPerBea
...
+	if (first.isHit[beamIdx]) {
+		first.xyz[beamIdx] = beamSamples.xyz[beamIdx * samplesPerBeam + firstIdx];
+		first.distance[beamIdx] = beamSamples.distance[beamIdx * samplesPerBeam + firstIdx];
+	}
+
+	last.isHit[beamIdx] = lastIdx > 0;
```
nebraszka, line 253: DITTO

src/gpu/nodeKernels.cu
```diff
@@ -229,25 +229,32 @@ __global__ void kProcessBeamSamplesFirstLast(size_t beamCount, int samplesPerBea
...
-		if (beamSamples.distance[beamIdx * samplesPerBeam + sampleIdx] <
-		    beamSamples.distance[beamIdx * samplesPerBeam + firstIdx]) {
+		auto currentFirstDistance = firstIdx >= 0 ? beamSamples.distance[beamIdx * samplesPerBeam + firstIdx] : FLT_MAX;
+		auto currentLastDistance = lastIdx >= 0 ? beamSamples.distance[beamIdx * samplesPerBeam + lastIdx] : -FLT_MAX;
+		if (beamSamples.distance[beamIdx * samplesPerBeam + sampleIdx] < currentFirstDistance) {
 			firstIdx = sampleIdx;
 		}
-		if (beamSamples.distance[beamIdx * samplesPerBeam + sampleIdx] >
-		    beamSamples.distance[beamIdx * samplesPerBeam + lastIdx]) {
+		if (beamSamples.distance[beamIdx * samplesPerBeam + sampleIdx] > currentLastDistance) {
```
PawelLiberadzki, line 243: Technically this if-statement and the above one exclude themselves - else-if could result is slightly better code, but it would only work assuming that we have current first and last distances already set. This can be done setting firstIdx and lastIdx (lines 232-233) to 0, and skipping sample at index 0 in for-loop in line 234.

src/gpu/nodeKernels.cu
```diff
@@ -229,25 +229,32 @@ __global__ void kProcessBeamSamplesFirstLast(size_t beamCount, int samplesPerBea
...
+	if (first.isHit[beamIdx]) {
+		first.xyz[beamIdx] = beamSamples.xyz[beamIdx * samplesPerBeam + firstIdx];
+		first.distance[beamIdx] = beamSamples.distance[beamIdx * samplesPerBeam + firstIdx];
+	}
+
+	last.isHit[beamIdx] = lastIdx >= 0;
+	if (last.isHit[beamIdx]) {
+		last.xyz[beamIdx] = beamSamples.xyz[beamIdx * samplesPerBeam + lastIdx];
+		last.distance[beamIdx] = beamSamples.distance[beamIdx * samplesPerBeam + lastIdx];
+	}
```
PawelLiberadzki, line 257: Is it not like if `beamSamples.isHit[beamIdx * samplesPerBeam + sampleIdx] != 0` (line 235 above) then you will always have first and last? If this is true, then two if-statements here may technically be reduced to one.

src/gpu/nodeKernels.cu
```diff
@@ -229,25 +229,32 @@ __global__ void kProcessBeamSamplesFirstLast(size_t beamCount, int samplesPerBea
...
 		}
 	}
-	first.xyz[beamIdx] = beamSamples.xyz[beamIdx * samplesPerBeam + firstIdx];
-	first.distance[beamIdx] = beamSamples.distance[beamIdx * samplesPerBeam + firstIdx];
-	last.xyz[beamIdx] = beamSamples.xyz[beamIdx * samplesPerBeam + lastIdx];
-	last.distance[beamIdx] = beamSamples.distance[beamIdx * samplesPerBeam + lastIdx];
+	first.isHit[beamIdx] = firstIdx >= 0;
+	if (first.isHit[beamIdx]) {
+		first.xyz[beamIdx] = beamSamples.xyz[beamIdx * samplesPerBeam + firstIdx];
+		first.distance[beamIdx] = beamSamples.distance[beamIdx * samplesPerBeam + firstIdx];
```
PawelLiberadzki, line 250: Regarding our discussion from today (setting both first and last at the main ray, not on side rays) are these xyz here in LiDAR space or in world? If this is a local space, and if I can assume that main ray is at index 0 between samplesPerBeam, then it could be possible to scale xyz of main ray hit point using its own distance and the distances of first and last.

PawelLiberadzki: On the other hand, adding self-transform to MultiReturnPointers (if obviously possible at all), would solve the case for world space. I guess, it is something to maybe discuss further.

src/graph/NodesCore.hpp
```diff
@@ -189,6 +191,33 @@ struct RaytraceNode : IPointsNode
...
+	// Node
+	void validateImpl() override
+	{
+		IPointsNodeSingleInput::validateImpl();
+		rtxInput = getExactlyOneInputOfType<RaytraceNode>();
+	}
+
+	void enqueueExecImpl() override {}
+
+	// Point cloud description
```
PawelLiberadzki, line 208: Probably unsed docs from copy/paste?

src/gpu/nodeKernels.cu
prybicki: Unfortunately, `xyz` is in world CF here.

> On the other hand, adding self-transform to MultiReturnPointers (if obviously possible at all), would solve the case for world space.

I think this could work, however cleanliness of such solution is very questionable, since it looses precision by converting back-and-forth. Let's wait and see if it's required to implement.

PawelLiberadzki (approved):
:1st_place_medal:

PR #287, opened 2024-05-20
Use c++ 11 only in api

AlgoryxJosef:
This change ensures applications built with C++11 will still be able to use rgl (public API), even if rgl itself uses newer c++ features internally. 
This is helpful when building for older Ubuntu versions for example. 

`std::is_trivial<rgl_vec2f>::value` is C++11, and `std::is_trivial_v<rgl_vec2f>` is C++17.

### DIFF
--- a/include/rgl/api/core.h
+++ b/include/rgl/api/core.h
@@ -71,8 +71,8 @@ typedef struct
 
 #ifdef __cplusplus
 static_assert(sizeof(rgl_vec2f) == 2 * sizeof(float));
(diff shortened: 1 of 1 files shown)


prybicki (approved):
Approved. However, without a test, it might be broken unknowingly in a day or two. Consider adding a test like this:
```
#if defined(__cplusplus) && (__cplusplus == 201103L)
#include "rgs/api/core.h"
#else
#pragma message("Error: test is supposed to be compiled as C++11")
#endif
```
Add it as separate target and set CXX_STANDARD property on it to 11.

PR #291, opened 2024-05-24
Add support for Laser Retro

msz-rai:
PR adds support for Laser Retro (floating-point value assigned to the entity and retrieved when hitting them).

The code is based on https://github.com/botsandus/RobotecGPULidar/pull/1 and adapted to the current version of the RGL.

### DIFF
--- a/include/rgl/api/core.h
+++ b/include/rgl/api/core.h
@@ -308,12 +308,6 @@ typedef enum : int32_t
 {
 	RGL_FIELD_XYZ_VEC3_F32 = 1,
 	RGL_FIELD_INTENSITY_F32,
-	/**
-	 * Lidar reflective value. Similar to the `RGL_FIELD_INTENSITY_F32` but set as a single value for the entire entity.
-	 * Could be replaced with `RGL_FIELD_INTENSITY_F32` and a 1x1 texture when float-type texture will be supported.
-	 * For non-hit points zero is assigned.
-	 */
-	RGL_FIELD_LASER_RETRO_F32,
 	RGL_FIELD_IS_HIT_I32,
 	RGL_FIELD_IS_GROUND_I32,
 	RGL_FIELD_RAY_IDX_U32,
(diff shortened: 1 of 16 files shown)


include/rgl/api/core.h
```diff
@@ -526,6 +532,16 @@ RGL_API rgl_status_t rgl_entity_set_id(rgl_entity_t entity, int32_t id);
...
 
+/**
+ * Set laser retro value for the given Entity.
+ * The value can be retrieved from `RGL_FIELD_LASER_RETRO_F32` point cloud field.
+ * Default retro for the Entity is zero.
+ * @param entity Entity to modify.
+ * @param retro Laser retro value to set.
+ */
+RGL_API rgl_status_t
+rgl_entity_set_laser_retro(rgl_entity_t entity, float retro);
```
PawelLiberadzki, line 543: Formatting is different than in other API calls - `RGL_API rgl_status_t` is one line before call name.

src/gpu/optixPrograms.cu
```diff
@@ -112,7 +115,7 @@ __forceinline__ __device__ void saveNonHitRayResult(float nonHitDistance)
 	displacement = {isnan(displacement.x()) ? 0 : displacement.x(), isnan(displacement.y()) ? 0 : displacement.y(),
 	                isnan(displacement.z()) ? 0 : displacement.z()};
 	Vec3f xyz = origin + displacement;
-	saveRayResult<false>(xyz, nonHitDistance, 0, RGL_ENTITY_INVALID_ID, Vec3f{NAN}, Vec3f{NAN}, 0.001f, Vec3f{NAN}, NAN);
+	saveRayResult<false>(xyz, nonHitDistance, 0, RGL_ENTITY_INVALID_ID, Vec3f{NAN}, Vec3f{NAN}, 0.0f, Vec3f{NAN}, NAN, 0.0f);
```
PawelLiberadzki, line 118: Nice catch ;) Is there a reason that one before last parameter (incidentAngle) is set to NAN here, when other parameters of this type are set to 0.0f.

PawelLiberadzki (approved):

src/gpu/optixPrograms.cu
msz-rai: Each field has been implemented individually and has its specifications to the non-hit value. We do not have a unification of non-hits by type. However, this is a good observation to consider in the future.

PawelLiberadzki:
LGTM! :)

PR #295, opened 2024-06-04
Compute XYZ in MR based on distance, not XYZ samples

prybicki:


src/gpu/optixPrograms.cu
```diff
@@ -186,7 +188,6 @@ extern "C" __global__ void __closesthit__()
 
 	// Save sub-sampling results
 	ctx.mrSamples.isHit[mrSamplesIdx] = true;
-	ctx.mrSamples.xyz[mrSamplesIdx] = hitWorldSeenBySensor;
```
PawelLiberadzki, line 189: Just to be sure that this was intended - is this not necessary because we simply sample distance alongside center ray to set xyz of first and last hit?

PawelLiberadzki (approved):
Nice updates in RaytraceNode and NodesCore!

src/gpu/optixPrograms.cu
prybicki: Yes, we don't need to assign here, since we compute XYZ later based on sampled distance.

PR #306, opened 2024-06-11
Revert extensions version (develop)

msz-rai:


prybicki (approved):
Very good changes :+1:

PR #308, opened 2024-06-13
Prevent underflow in loop condition

nebraszka:


msz-rai (approved):
Nice catch, good job :1st_place_medal:

Issue #313, opened 2024-07-11
Pointcloud message type incompatibility with Autoware


Aysenayilmaz:
With https://github.com/autowarefoundation/autoware.universe/pull/6996, the pointcloud types have been changed. They have transitioned from using _PointXYZI_ to _PointXYZIRCAEDT_. 

```
- X: x   
- Y: y
- z: z
- I: intensity
- R: ring
- A :azimuth
- D: distance
- T: time_stamp
```
As a result, pointcloud types are no longer supported by the RobotecGPULidar.
### Current point types:  
```
- 'x'  FLOAT32
- 'y' FLOAT32
- 'z'  FLOAT32
- 'intensity'  FLOAT32
- 'ring' UINT16
- 'azimuth'  FLOAT32
- 'distance' FLOAT32
- 'return_type' UINT8
- 'time_stamp' FLOAT64
```
### Data types used in Autoware:   
https://github.com/autowarefoundation/autoware.universe/blob/9f41bf13d9e8a73d2bfd6278cc249e2306348c1e/common/autoware_point_types/include/autoware_point_types/types.hpp#L95
```
- x,y,z  float   	 
- intensity  uint8_t   	 
- return_type  uint8_t   	 
- channel  uint16_t    
- azimuth  float   	 
- elvation   float   	 
- distance  float   	 
- time_stamp   uint32_t 
```   
That's why there is a https://github.com/autowarefoundation/AWSIM-Labs/issues/114 problem in AWsim Labs. 
The error I get when I run Autoware and AWsim Labs:
`[component_container-30] [ERROR] [1720621336.877892482] [sensing.lidar.top.crop_box_filter_self]: The pointcloud layout is not compatible with PointXYZIRCAEDT/PointXYZIRC. Aborting`

prybicki:
@Aysenayilmaz This is not an RGL issue. Pointcloud format in RGL is configurable in runtime using `rgl_node_points_format`. Therefore I'd recommend to have a look at the code that's using RGL, most likely [here](https://github.com/autowarefoundation/AWSIM-Labs/tree/main/Assets/AWSIM/Scripts/Sensors/LiDAR).

msz-rai:
@Aysenayilmaz 

We initially underestimated the issue, failing to recognize that the field types had also changed. This necessitates the implementation of data type conversions in the native library. We are addressing this and will implement the necessary changes as soon as possible.

Aysenayilmaz:
[...]

PR #315, opened 2024-07-17
Add (some) alternative field types to support new Autoware format

msz-rai:
This PR:
- adds new fields: `RGL_FIELD_INTENSITY_U8` and `RGL_FIELD_TIME_STAMP_U32`
  - these fields are duplicates of existing fields but are stored in other data types
- changes what time stamp fields hold
  - time passed since the time of the sensor trigger when the given point was measured instead of scene time
  - it is consistent with the Autoware convention now
- renames ROS2 field name for RGL_FIELD_RING_ID:
  - `ring` -> `channel`


src/RGLFields.hpp
```diff
@@ -318,20 +332,25 @@ inline std::vector<uint8_t> toRos2Fields(rgl_field_t type)
...
 	switch (type) {
 		case XYZ_VEC3_F32: return {"x", "y", "z"};
 		case IS_HIT_I32: return {"is_hit"};
 		case IS_GROUND_I32: return {"is_ground"};
 		case ENTITY_ID_I32: return {"entity_id"};
 		case RAY_IDX_U32: return {"ray_idx"};
-		case INTENSITY_F32: return {"intensity"};
+		case INTENSITY_F32:
+		case INTENSITY_U8:
+			return {"intensity"};
```
PawelLiberadzki, line 344: Because of these fields sharing a name, can there be a niche case somewhere (ROS 2 publisher?) that will cause problems in case of both intensities requested?

PawelLiberadzki (approved):
Looks great! I had only one minor comment.

src/RGLFields.hpp
msz-rai: This function is used by ROS2 publishers only and I see no blocker for having two fields with the same name.

Aysenayilmaz:
Related issue : 

-  https://github.com/RobotecAI/RobotecGPULidar/issues/313

PR #317, opened 2024-07-19
Consider incident angle when calculating intensity

msz-rai:
This PR:
- takes into account the influence of the incident angle on the intensity value
- adds API call for setting default intensity.
  - intensity considered when hitting entities with no intensity texture set

### DIFF
--- a/src/gpu/RaytraceRequestContext.hpp
+++ b/src/gpu/RaytraceRequestContext.hpp
@@ -28,6 +28,8 @@ struct RaytraceRequestContext
 	float nearNonHitDistance;
 	float farNonHitDistance;
 
+	float defaultIntensity;
+
 	const Mat3x4f* raysWorld;
 	size_t rayCount;
 
--- a/src/gpu/optixPrograms.cu
+++ b/src/gpu/optixPrograms.cu
@@ -188,7 +188,7 @@ extern "C" __global__ void __closesthit__()
 	const Vec3f wNormal = wAB.cross(wCA).normalized();
 	const float incidentAngle = acosf(fabs(wNormal.dot(rayDir)));
 
-	float intensity = 0;
+	float intensity = ctx.defaultIntensity;
 	bool isIntensityRequested = ctx.intensityF32 != nullptr || ctx.intensityU8 != nullptr;
 	if (isIntensityRequested && entityData.textureCoords != nullptr && entityData.texture != 0) {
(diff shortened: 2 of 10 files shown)


src/gpu/RaytraceRequestContext.hpp
```diff
@@ -28,6 +28,8 @@ struct RaytraceRequestContext
 	float nearNonHitDistance;
 	float farNonHitDistance;
 
+	float defaultIntensity;
```
PawelLiberadzki, line 31: Do we agree on default here being 0.0f?

scorp18pl: Wouldn't 0.5f make more sense?

alek-kam-robotec-ai: Wouldn't 0.5f make more sense?

PawelLiberadzki (approved):

src/gpu/optixPrograms.cu
```diff
@@ -203,6 +203,7 @@ extern "C" __global__ void __closesthit__()
 
 		intensity = tex2D<TextureTexelFormat>(entityData.texture, uv[0], uv[1]);
 	}
+	intensity *= cosf(incidentAngle);
```
scorp18pl, line 206: @msz-rai  Isn't this redundant? The absolute value of the dot product is already known, no need to calculate cos on an acos output. The above incident angle calculation has to be simply split into two statements. One to calculate and store the absolute value of the dot product (this value can be used here). Second one for calculation of the incident angle. To the best of my knowledge tryg functions are rather expensive (i would guess that even more so on the gpu).

src/gpu/optixPrograms.cu
```diff
@@ -188,7 +188,7 @@ extern "C" __global__ void __closesthit__()
 	const Vec3f wNormal = wAB.cross(wCA).normalized();
 	const float incidentAngle = acosf(fabs(wNormal.dot(rayDir)));
```
alek-kam-robotec-ai, line 189: The line I'm suggesting to split.

src/gpu/RaytraceRequestContext.hpp
msz-rai: Discussed offline

src/gpu/optixPrograms.cu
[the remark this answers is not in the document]
msz-rai: Good point. :+1:

PR #318, opened 2024-07-23
v17 O3DE integration hotfix

alek-kam-robotec-ai:
**This pr:**
  - Moves the `deprecated` attribute before the `RGL_API` macro, which previously failed due to a compiler error.
  - Adds additional installation instruction related to the `radar-msgs` ros2 package dependency.

### DIFF
--- a/include/rgl/api/core.h
+++ b/include/rgl/api/core.h
@@ -809,16 +809,6 @@ RGL_API rgl_status_t rgl_node_points_format(rgl_node_t* node, const rgl_field_t*
 RGL_API rgl_status_t rgl_node_points_yield(rgl_node_t* node, const rgl_field_t* fields, int32_t field_count);
 
 /**
- * Creates or modifies CompactPointsNode.
- * The Node removes non-hit points. In other words, it converts a point cloud into a dense one.
- * Graph input: point cloud
- * Graph output: point cloud (compacted)
- * @param node If (*node) == nullptr, a new Node will be created. Otherwise, (*node) will be modified.
- */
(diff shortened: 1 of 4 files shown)


include/rgl/api/core.h
```diff
@@ -741,7 +741,7 @@ RGL_API rgl_status_t rgl_node_points_yield(rgl_node_t* node, const rgl_field_t*
  * Graph output: point cloud (compacted)
  * @param node If (*node) == nullptr, a new Node will be created. Otherwise, (*node) will be modified.
  */
-RGL_API [[deprecated("Use rgl_node_points_compact_by_field(rgl_node_t* node, rgl_field_t field) instead.")]] rgl_status_t
+[[deprecated("Use rgl_node_points_compact_by_field(rgl_node_t* node, rgl_field_t field) instead.")]] RGL_API rgl_status_t
```
PawelLiberadzki, line 744: This is the reason behind CI failing - extern "C" (being under RGL_API macro) can only appear at the start of a declaration.

PawelLiberadzki: Here is a discussion on similar topic:
https://github.com/llvm/llvm-project/issues/56077

alek-kam-robotec-ai: I see, but compiling the current version of "core.h" under o3de fails with a compiler error:
```bash
rgl/api/core.h:818:11: error: 'deprecated' attribute cannot be applied to types
RGL_API [[deprecated("Use rgl_node_points_compact_by_field(rgl_node_t* node, rgl_field_t field) instead.")]] rgl_status_t
          ^
1 error generated.
ninja: build stopped: subcommand failed.
```
I'm not sure if this code ever worked if compiled under a c++ compiler. I believe that tests only cover the library internals.

msz-rai (approved):

PR #324, opened 2024-08-02
Update multi-return for more types of return and all RGL fields.

PawelLiberadzki:
### Description

This PRs introduces an update to RGL multi-return - more types are handled now (first, second, last, strongest and second strongest) and all RGL fields are handled. Moreover, this vastly simplifies an amount of effort to introduce multi-return for ROS 2 and UDP publishing.

A temporary downgrade is that multi-return samples are allocated even if beam divergence is set to zero. However, that was the case also before this update. This will be solved in close future PRs.

Multi-return modes and beam divergences are configurable at runtime.

### Testing

Run implemented unit tests - most important here is asyncStressTest. For runtime check I propose running cube_in_motion test from multiReturnTest suite - uncomment GTEST_SKIP there. Results are visible in RViz2 ("world" frame). Divergence is configured on beams, and return modes switch circularly in following order:
 - RGL_RETURN_FIRST
 - RGL_RETURN_LAST
 - RGL_RETURN_FIRST_LAST

### DIFF
--- a/include/rgl/api/core.h
+++ b/include/rgl/api/core.h
@@ -816,6 +816,18 @@ RGL_API rgl_status_t rgl_node_raytrace_configure_beam_divergence(rgl_node_t node
 RGL_API rgl_status_t rgl_node_raytrace_configure_default_intensity(rgl_node_t node, float default_intensity);
 
 /**
+ * Modifies RaytraceNode to set return mode.
+ * Point return types (RGL_FIELD_RETURN_TYPE_U8) will be set to corresponding rgl_return_type_t values, e.g. return mode
+ * RGL_RETURN_FIRST_LAST will result in point return types to be set to RGL_RETURN_TYPE_FIRST or RGL_RETURN_TYPE_LAST,
+ * interchangeably.
+ * Default return mode on RaytraceNode is RGL_RETURN_FIRST.
+ * Passing RGL_RETURN_MODE_UNDEFINED will have no effect (does not change return mode).
+ * @param node RaytraceNode to modify.
+ * @param return_mode Return mode to set.
+ */
+RGL_API rgl_status_t rgl_node_raytrace_configure_return_mode(rgl_node_t node, rgl_return_mode_t return_mode);
+
+/**
  * Creates or modifies FormatPointsNode.
  * The Node converts internal representation into a binary format defined by the `fields` array.
  * Note: It is the user's responsibility to ensure proper data structure alignment. See (https://en.wikipedia.org/wiki/Data_structure_alignment).
--- a/include/rgl/api/core.h
+++ b/include/rgl/api/core.h
@@ -1068,17 +1068,6 @@ RGL_API rgl_status_t rgl_node_gaussian_noise_distance(rgl_node_t* node, float me
                                                       float st_dev_rise_per_meter);
 
 /**
- * Creates or modifies MultiReturnSwitchNode
- * This is a special node which does not modify the data but acts as an adapter to the multi-return feature.
- * Thanks to this node, user can attach unchanged pipelines to work with specific return type from multi-return raytracing.
- * Graph input: point cloud (with multi-return fields)
- * Graph output: point cloud (with a selected field from parent's multi-return point cloud)
- * @param node If (*node) == nullptr, a new Node will be created. Otherwise, (*node) will be modified.
- * @param return_type Return type to select from multi-return point cloud.
- */
-RGL_API rgl_status_t rgl_node_multi_return_switch(rgl_node_t* node, rgl_return_type_t);
-
-/**
  * Assigns value true to out_alive if the given node is known and has not been destroyed,
  * assigns value false otherwise.
  * @param node Node to check if alive
--- a/include/rgl/api/core.h
+++ b/include/rgl/api/core.h
@@ -423,43 +423,57 @@ typedef enum : int32_t
 } rgl_field_t;
 
 /**
+ * Bitshift and masks for coding return modes and types. Most significant byte encodes the number of returns. Return mode is
+ * encoded on 8 most significant bits (24-31) - 1 for single return mode, 2 for dual return. Remaining 24 bits are for encoding
+ * return type. First return is on 8 LEAST significant bits (0-7), and next 8 bits (8-15) are for second return. Bits 16-23 are
+ * not used at this moment - in the future e.g. the third return may be encoded there. Return type comes from rgl_return_type_t.
+ * Example: RGL_RETURN_LAST_STRONGEST would be encoded as:
+ * 		0x02000201
+ * 			02 - dual return
+ * 			00 - not used bits
+ * 			02 - second return - RGL_RETURN_TYPE_LAST
+ *			01 - first return - RGL_RETURN_TYPE_STRONGEST
+ */
+const int32_t RGL_RETURN_MODE_BIT_SHIFT = 24;
+const int32_t RGL_RETURN_MODE_SINGLE = 1 << RGL_RETURN_MODE_BIT_SHIFT;
+const int32_t RGL_RETURN_MODE_DUAL = 2 << RGL_RETURN_MODE_BIT_SHIFT;
+const int32_t RGL_RETURN_TYPE_BIT_SHIFT = 8;
+
+/**
  * Kinds of return types for multi-return LiDAR output.
  */
 typedef enum : uint8_t
 {
-	RGL_RETURN_TYPE_UNKNOWN = 1 << 0,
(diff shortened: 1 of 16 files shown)


src/gpu/nodeKernels.cu
```diff
@@ -223,41 +334,97 @@ __global__ void kFilterGroundPoints(size_t pointCount, const Vec3f sensor_up_vec
...
+	for (sampleIdx = sampleIdx + 1; sampleIdx < firstSampleInBeamIdx + samplesPerBeam; ++sampleIdx) {
+		if (ctx->mrSamples.isHit[sampleIdx] == 0) {
 			continue;
 		}
-		auto currentFirstDistance = firstIdx >= 0 ? beamSamples.distance[beamIdx * samplesPerBeam + firstIdx] : FLT_MAX;
-		auto currentLastDistance = lastIdx >= 0 ? beamSamples.distance[beamIdx * samplesPerBeam + lastIdx] : -FLT_MAX;
-		if (beamSamples.distance[beamIdx * samplesPerBeam + sampleIdx] < currentFirstDistance) {
-			firstIdx = sampleIdx;
+
+		const auto currentSampleDistance = ctx->mrSamples.distance[sampleIdx];
```
msz-rai, line 377: `currentSampleDistance` -> `currentDistance` for consistency with `currentIntensity` defined later.

msz-rai (approved):
Good job! :mechanical_arm: 
Minor suggestions to consider.

src/gpu/optixPrograms.cu
msz-rai, line 1: Intensity value should be retrieved from the intensity texture if the strongest return type is requested but the intensity field is not. The easiest way will always be to take the intensity texture.

src/gpu/nodeKernels.cu
```diff
@@ -100,6 +100,117 @@ __device__ Vec3f reflectPolarization(const Vec3f& pol, const Vec3f& hitNormal, c
...
+		const Mat3x4f ray = ctx->raysWorld[beamIdx];
+		const Vec3f origin = ray * Vec3f{0, 0, 0};
+		const Vec3f dir = (ray * Vec3f{0, 0, 1} - origin).normalized();
+		ctx->xyz[returnPointIdx] = origin + dir * ctx->mrSamples.distance[sampleIdx];
+	}
+	if (ctx->isHit != nullptr) {
+		ctx->isHit[returnPointIdx] = ctx->mrSamples.isHit[sampleIdx];
+	}
+	if (ctx->ringIdx != nullptr && ctx->ringIds != nullptr) {
+		ctx->ringIdx[returnPointIdx] = ctx->ringIds[beamIdx % ctx->ringIdsCount];
```
msz-rai, line 116: Probably, `ringIdx` can be filled in `saveBeamSharedData()` regardless of hit or non-hit.

prybicki (approved):
LGTM

include/rgl/api/core.h
```diff
@@ -785,6 +829,18 @@ RGL_API rgl_status_t rgl_node_raytrace_configure_beam_divergence(rgl_node_t node
  */
 RGL_API rgl_status_t rgl_node_raytrace_configure_default_intensity(rgl_node_t node, float default_intensity);
 
+/**
+ * Modifies RaytraceNode to set return mode.
+ * Point return types (RGL_FIELD_RETURN_TYPE_U8) will be set to corresponding rgl_return_type_t values, e.g. return mode
+ * RGL_RETURN_FIRST_LAST will result in point return types to be set to RGL_RETURN_TYPE_FIRST or RGL_RETURN_TYPE_LAST,
+ * interchangeably.
+ * Default return mode on RaytraceNode is RGL_RETURN_FIRST.
+ * Passing RGL_RETURN_MODE_UNDEFINED will have no effect (does not change return mode).
```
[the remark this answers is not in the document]
PawelLiberadzki: Correct - this is the last place that I must have missed. I add additional explanation.

src/api/apiCore.cpp
```diff
@@ -1002,6 +1002,28 @@ void TapeCore::tape_node_raytrace_configure_default_intensity(const YAML::Node&
...
+	return status;
+}
+
+void TapeCore::tape_node_raytrace_configure_return_mode(const YAML::Node& yamlNode, PlaybackState& state)
+{
+	auto nodeId = yamlNode[0].as<TapeAPIObjectID>();
+	auto returnMode = static_cast<rgl_return_mode_t>(yamlNode[1].as<int>());
+	rgl_node_t node = state.nodes.contains(nodeId) ? state.nodes.at(nodeId) : nullptr;
+	rgl_node_raytrace_configure_return_mode(node, returnMode);
+	state.nodes.insert({nodeId, node});
```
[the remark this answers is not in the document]
PawelLiberadzki: Nice catch!

src/gpu/optixPrograms.cu
```diff
@@ -154,27 +124,11 @@ extern "C" __global__ void __closesthit__()
 	const Vector<3, double> hwrd = hitWorldRaytraced;
 	const Vector<3, double> hso = beamSampleOrigin;
 	const double distance = (hwrd - hso).length();
```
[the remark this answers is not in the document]
PawelLiberadzki: Not mine. However I may only guess this is because of some precision issues when calculating distance. Especially when you look at two vectors above, that are explicitly made double too. I would rather leave it, but I can also change it - what do you think? I casted distance to float, leaving double for length calculation and it seems to work - tests pass.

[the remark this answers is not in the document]
PawelLiberadzki: Discussed offline. It was a precision issue at some point in the past. Does not seem to happen now, however going back to float would require some tests. I put a TODO here for now.

src/gpu/nodeKernels.cu
```diff
@@ -223,41 +334,97 @@ __global__ void kFilterGroundPoints(size_t pointCount, const Vec3f sensor_up_vec
...
+			first = sampleIdx;
+			second = first;
+			last = first;
+			strongest = sampleIdx;
+			secondStrongest = strongest;
+			break;
+		}
+	}
+
+	const auto returnCount = returnMode >> RGL_RETURN_MODE_BIT_SHIFT;
```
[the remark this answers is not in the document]
PawelLiberadzki: I added this and also introduced `getReturnType()` and updated it in nodeKernels.

src/gpu/nodeKernels.cu
```diff
@@ -223,41 +334,97 @@ __global__ void kFilterGroundPoints(size_t pointCount, const Vec3f sensor_up_vec
...
+
+	// There were no hits within beam samples.
+	if (first == -1) {
+		for (int returnIdx = 0; returnIdx < returnCount; ++returnIdx) {
+			const auto returnPointIdx = beamIdx * returnCount + returnIdx;
+			const auto returnType = (returnMode >> (returnIdx * RGL_RETURN_TYPE_BIT_SHIFT)) & 0xff;
```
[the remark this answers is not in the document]
PawelLiberadzki: Done - please check returnModeUtils.h.

src/gpu/optixPrograms.cu
PawelLiberadzki: Updated - now texture is handled always if it is set, i.a. if user sets texture, sets return mode based on intensity (containing strongest or second strongest), but does not request intensity field then return will consider this texture.

I also left TODO here, because it should be possible to depend on return mode in this place. I will try to optimize this in next step.

src/gpu/nodeKernels.cu
PawelLiberadzki: I updated code like that.

src/returnModeUtils.h
```diff
@@ -14,9 +14,42 @@
...
+		{RGL_RETURN_TYPE_UNKNOWN, "Unknown"},
+		{RGL_RETURN_TYPE_FIRST, "First"},
+		{RGL_RETURN_TYPE_SECOND, "Second"},
+		{RGL_RETURN_TYPE_LAST, "Last"},
+	    {RGL_RETURN_TYPE_STRONGEST, "Strongest"},
+	    {RGL_RETURN_TYPE_SECOND_STRONGEST, "SecondStrongest"},
+	};
+	// clang-format on
+	assert(typeNames.find(returnType) != typeNames.cend());
+	return typeNames.at(returnType);
```
msz-rai, line 45: `switch` may be cleaner for this, but I am not forcing it.

PawelLiberadzki: I will stay with a map here, that on the other hand feels cleaner for me. And I see no downgrade of using map.

PR #332, opened 2024-09-05
Fix: use first sample of given beam when calculating output for non-hits

msz-rai:
Before, the first sample of the first beam was used for all output points


PawelLiberadzki (approved):
Great catch!

PR #333, opened 2024-09-25
Minor fixes around yield and compact node

msz-rai:
- Fix yield node to return hasField() according to its map
- Require a field that is actually used by the compact node


PawelLiberadzki (approved):
Nice catch, looks great!

PR #339, opened 2024-10-17
Fix hit point displacement calculation resulting from animation

msz-rai:


PawelLiberadzki (approved):
Looks nice, great catch!

PR #347, opened 2025-03-31
Add support for ignoring entities in sensor raytracing

PiotrMrozik:


include/rgl/api/core.h
```diff
@@ -672,6 +680,13 @@ RGL_API rgl_status_t rgl_entity_set_pose_world(rgl_entity_t entity, const rgl_ma
  */
 RGL_API rgl_status_t rgl_entity_set_id(rgl_entity_t entity, int32_t id);
 
+/**
+ * Set sensor ID of the given Entity. This ID is used to identify the sensor in the scene and omit hitpoints on objects with the same ID.
+ * @param entity Entity to modify
+ * @param sensor_id ID to set. If not set, value of the Entity sensor id will be automatically generated as a DEFAULT_SENSOR_ID.
+ */
+RGL_API rgl_status_t rgl_entity_set_sensor_id(rgl_entity_t entity, int32_t sensor_id);
```
msz-rai: Naming suggestion: `rgl_entity_set_ignored_by_sensor`

include/rgl/api/core.h
```diff
@@ -61,6 +61,14 @@
 // It is assigned by default if the user does not specify it.
 #define RGL_DEFAULT_ENTITY_ID 268435455
 
+// Default Entity-Sensor ID is the largest positive 28-bit integer (OptiX limit).
```
msz-rai: Since it is used in `RaytraceRequestContext` and `EntitySBTData` as an int there is no `OptiX limit` to 28-bit.

include/rgl/api/core.h
```diff
@@ -844,6 +859,15 @@ RGL_API rgl_status_t rgl_node_raytrace(rgl_node_t* node, rgl_scene_t scene);
 RGL_API rgl_status_t rgl_node_raytrace_configure_velocity(rgl_node_t node, const rgl_vec3f* linear_velocity,
                                                           const rgl_vec3f* angular_velocity);
 
+/**
+ * Modifies RaytraceNode to apply sensor identifier.
+ * This id will correlate with the entities ids in the scene. Entities with the same id as the sensor will be ignored.
+ * Can be used to ignore the sensor itself in the raytracing.
+ * @param node RaytraceNode to modify
+ * @param id Sensor identifier to set. Default value is 0 and will not ignore any entity.
```
msz-rai, line 864: You define `RGL_DEFAULT_SENSOR_ID` as -1 above, but here you refer to it as 0.

src/gpu/ShaderBindingTableTypes.h
```diff
@@ -19,6 +19,7 @@ struct EntitySBTData
 	size_t textureCoordsCount;
 	cudaTextureObject_t texture;
 	float laserRetro;
+	int entitySensorId;
```
msz-rai: Naming suggestion: `ignoredBySensorId`

src/gpu/optixPrograms.cu
```diff
@@ -112,10 +112,12 @@ extern "C" __global__ void __closesthit__()
 
 	// Ray
 	const int beamIdx = static_cast<int>(optixGetLaunchIndex().x);
-	const int beamSampleRayIdx = static_cast<int>(optixGetPayload_0());
-	const int mrSampleIdx = beamIdx * MULTI_RETURN_BEAM_SAMPLES + beamSampleRayIdx;
+	unsigned int beamSampleRayIdx = optixGetPayload_0();
+	unsigned int mrSampleIdx = beamIdx * MULTI_RETURN_BEAM_SAMPLES + beamSampleRayIdx;
```
msz-rai: I believe they can be still const.

src/gpu/optixPrograms.cu
```diff
@@ -135,6 +137,18 @@ extern "C" __global__ void __closesthit__()
 		return;
 	}
 
+	// Ignore entities with corresponding IDs.
+	if (entitySensorID == sensorId) {
+		constexpr auto epsilon = 1.19209e-06f;
```
msz-rai: I'm curious where this precise value of epsilon came from. I'd expect a round number (e.g. `1e-6f`).
Also, please clarify in the comment why this epsilon is necessary. I assume it must be applied to avoid hitting the same mesh (triangle) again, but I'm not sure.

src/gpu/optixPrograms.cu
```diff
@@ -135,6 +137,18 @@ extern "C" __global__ void __closesthit__()
...
+		const Vec3f dir = optixGetWorldRayDirection();
+		ctx.mrSamples.distance[mrSampleIdx] += distance + epsilon;
+		const float maxRange = ctx.rayRangesCount == 1 ? ctx.rayRanges[0].y() : ctx.rayRanges[mrSampleIdx].y();
+
+		optixTrace(ctx.scene, hitWorldRaytraced + epsilon * dir, dir, epsilon, maxRange, 0.0f, OptixVisibilityMask(255),
+			OPTIX_RAY_FLAG_NONE, 0, 1, 0, beamSampleRayIdx);
```
msz-rai: 1. Setting `epsilon` as a `tmin` parameter will cause it to be considered twice (first when evaluating new origin of the ray and second when setting `tmin`). Is this intentional?
2. When tracing rays for the first time we use `OPTIX_RAY_FLAG_DISABLE_ANYHIT` flag. Is changing the flag intentional?

src/scene/Entity.hpp
```diff
@@ -128,6 +133,8 @@ struct Entity : APIObject<Entity>
 	TransformWithTime formerTransformInfo{Mat3x4f::identity(), std::nullopt};
 
 	Field<ENTITY_ID_I32>::type id{RGL_DEFAULT_ENTITY_ID};
+	int sensorId{RGL_DEFAULT_ENTITY_SENSOR_ID};
```
msz-rai: Naming suggestion: `ignoredBySensorId` with default value `RGL_INVALID_SENSOR_ID`. Invalid means that id cannot be set for the sensor.

msz-rai (commented):
GJ!
Please address the comments.

src/scene/Entity.hpp
msz-rai: `RGL_SENSOR_ID_NONE` instead of `RGL_INVALID_SENSOR_ID` will also be a good match.

src/gpu/optixPrograms.cu
PiotrMrozik: `beamSampleRayIdx` unfortunately not. 
But anyway, good catch with `mrSampleIdx`

msz-rai (approved):

PR #351, opened 2025-05-28
Bump CUDA architecture (remove deprecated one)

msz-rai:


README.md
```diff
@@ -45,9 +45,9 @@ And more:
 
 ## Runtime requirements
 
-| Hardware | Requirement  |
-|:---------|:-------------|
-| GPU      | CUDA-enabled |
+| Hardware | Requirement                                   |
+|:---------|:----------------------------------------------|
+| GPU      | `>=Nvidia Turing` (GeForce RTX 2060 or newer) |
```
prybicki: I'd say `Nvidia Turing (GeForce RTX 2000 series) or newer`. E.g. 2080 is usually released **before** 2060.

CMakeLists.txt
```diff
@@ -188,7 +192,6 @@ target_link_libraries(RobotecGPULidar PUBLIC
     CUDA::nvml
     CUDA::cudart_static
     CUDA::cuda_driver
-    CUDA::nvToolsExt
```
prybicki, line 191: Does it mean nvtx markers no longer work? If so, maybe a comment what's needed to make it work?

prybicki (approved):

CMakeLists.txt
msz-rai: It (NVTX v2) has already been removed since CUDA 12.9 ([link](https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#id4)). However, we are already using NVTX v3, which is header-only library and no dependency is needed.

PR #356, opened 2025-08-18
Add detection states in radar nodes

PawelLiberadzki:
This PR adds cluster statistics (std for azimuth, elevation, distance, radial speed) to RadarPostprocessPointsNode. This statistics is added to previously available cluster AABBs, and exposed through Node API. This data is consumed further in RadarTrackObjectsNode in order to calculate detection states.


msz-rai (approved):
Solid implementation :mechanical_arm:

PR #364, opened 2026-03-16
Disable CI workflow

msz-rai:
The workflow has been disabled because the self-hosted computer has been taken out of service for this project.


alek-kam (approved):
👍

